PR #3, opened 2021-01-27
Add warning about Galera cluster bootstrapping

srueg:


docs/modules/ROOT/pages/how-tos/mariadbgalera/bootstrap_node.adoc
```diff
@@ -1,5 +1,10 @@
 = How to Bootstrap a Cluster (other than from Node 0)
 
+[WARNING]
+====
+If you bootstrap from ayn other node than the one that has proceeded the furthest in transactions, this will result in data loss.
```
tobru, line 5:
```suggestion
If you bootstrap from any other node than the one that has proceeded the furthest in transactions, this will result in data loss.
```

tobru (approved):

PR #10, opened 2021-12-08
Document Application Catalog Concept

tobru:


docs/modules/ROOT/pages/explanations/app_catalog.adoc
```diff
@@ -0,0 +1,118 @@
...
+
+It is our framework to fully automate service provisioning and all tasks for day-2 operations.
+
+image::appcat_concept.drawio.svg[]
+
+== Service Location
+
+Services can be provisioned on various places:
+
+Local Cluster::
```
Kidswiss, line 19: Maybe call it control cluster?

docs/modules/ROOT/pages/explanations/app_catalog.adoc
```diff
@@ -0,0 +1,118 @@
...
+Service runs on the same cluster as the control-plane.
+Either in the same namespace as the service request was made or in another namespace.
+Usually deployed via a Helm Chart.
+
+Remote (Service) Cluster::
```
Kidswiss, line 24: I'd just call it service cluster. Maybe we have potential deployments where the control and service clusters are the same. Then it's a bit confusing to talks about local and remote.

docs/modules/ROOT/pages/explanations/app_catalog.adoc
```diff
@@ -0,0 +1,118 @@
...
+These are our own custom services and are engineered and provided by VSHN.
+
+Configuration Provider::
+Some providers directly configure a service which was provisioned by another provider.
+For example https://github.com/crossplane-contrib/provider-sql[provider-sql] can configure users and permissions in an SQL database which was provisioned by the Helm Provider or by a Cloud Services Provider.
+
+== Configuration Management
+
+Crossplane itself is deployed and configured with the Project Syn https://github.com/projectsyn/component-crossplane[Commodore Component].
+Although there is the concept of https://crossplane.io/docs/master/concepts/packages.html[Crossplane packages], service definitions (XRDs and Compositions) and provider installation and configuration is delegated to Project Syn Commodore Components as well.
```
Kidswiss, line 77: Does that mean we imitate Crossplane packages via Syn or that we apply them via Syn?

docs/modules/ROOT/pages/explanations/app_catalog.adoc
```diff
@@ -0,0 +1,118 @@
...
+
+=== Functional
+
+* Provisioning of a service happens fully automated
+* Configuration of services is a declarative approach
+* Day-2 operations of a service is well defined and automated as far as possible
+
+=== Performance
+
+* Services are provisioned in under 1 minute
```
Kidswiss, line 92: Not sure if this is realistic :D Some applications take longer to initialise than that.

docs/modules/ROOT/pages/explanations/app_catalog.adoc
```diff
@@ -0,0 +1,118 @@
...
+* Every service defines its own performance requirements (answer time, reaction time, ...) and constraints
+
+=== Reliability
+
+* Services have a defined SLI (Service Level Indicator) and an appropriate SLO (Service Level Objective)
+* Backup of data is regularly conducted
```
Kidswiss, line 98: No word about if we monitor them or not. Could lead to wrong conclusions.

Kidswiss (changes requested):

docs/modules/ROOT/pages/explanations/app_catalog.adoc
tobru: I would distinguish between these names, as the control cluster doesn't necessarily need to run services.

docs/modules/ROOT/pages/explanations/app_catalog.adoc
tobru: What about "Dedicated Service Cluster"?

docs/modules/ROOT/pages/explanations/app_catalog.adoc
tobru: The idea is to imitate the Crossplane packages and not use Crossplane packages, it would add another layer of complication, as Commodore Components already do the same (deploy yaml files).

docs/modules/ROOT/pages/explanations/app_catalog.adoc
tobru: Yes, that's true... We anyways have to put more work in this whole section.

docs/modules/ROOT/pages/explanations/app_catalog.adoc
tobru: Yes, I'm unsure where and how to write that.

tobru:
@Kidswiss Thanks for the review. I'm thinking of moving the whole section "Quality of a Service" into its own PR. WDYT?

docs/modules/ROOT/pages/explanations/app_catalog.adoc
Kidswiss: I agree that distinguishing them is a good thing. But local and remote are a bit unclear.

Maybe something like this:
control cluster -> cluster dedicated to hosting the controlplane
service cluster -> cluster dedicated to running the services
converged cluster -> cluster running both and other customer services if neccassry.

Should help with ambiguities I think.

docs/modules/ROOT/pages/explanations/app_catalog.adoc
Kidswiss: See suggestion above.

docs/modules/ROOT/pages/explanations/app_catalog.adoc
Kidswiss: Fine by me.

Kidswiss:
@tobru Would make sense, because that's the part where we should be the most precise so the customer knows exactly what to expect.

tobru:
I introduced a terminology section and move quality requirements to its own PR:

* https://github.com/vshn/application-catalog-docs/pull/13

Kidswiss (approved):
LGTM

PR #19, opened 2022-03-17
Decision: Converged Service provisioning implementation

ccremer:
## Summary

This PR documents the general approach how services are provisioned using the Crossplane technology.

## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/explanations/decisions/crossplane-impl.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/crossplane-impl.adoc
@@ -13,7 +13,7 @@ See discussions in https://github.com/crossplane/crossplane/issues/1895[#1895] a
 Currently the https://github.com/mittwald/kubernetes-secret-generator[Secret generator] is used but this should be regarded as a workaround.
 
 Webhook support for validations and custom business logic::
-See design proposal in https://github.com/crossplane/crossplane/pull/2719[#2719].
+With release v1.7.0 Crossplane offers methods to build webhooks, but apparently only for dedicated Providers.
 
 Delay reconciliations to specific times when using `CompositionRevisions` feature::
 This is needed to enable maintenance windows.
@@ -21,6 +21,14 @@ This is needed to enable maintenance windows.
 Cross-resource references within `Composition`::
 This could eliminate a lot of patch duplications.
 See discussions in https://github.com/crossplane/crossplane/issues/1770[#1770] and https://github.com/crossplane/crossplane/issues/2099[#2099].
+
+More methods for dynamic value computation::
+It's currently impossible to apply certain types of patching:
++
+- Transformation of string to boolean with an expression evaluator, for example `if spec.sla == "guaranteed" then true else false`
+- Omit deploying resources, for example if `spec.backups.enabled` is set to `false` in the XR, it's impossible to skip generating a resource based on this conditional.
+- Basic math operations like addition, subtraction division or floating-point multiplication.
+- The `map` patch type only supports string-to-string mapping, not string-to-object which would help patching whole specs.
 
 [NOTE]
 ====
--- a/docs/modules/ROOT/pages/explanations/decisions/crossplane-impl.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/crossplane-impl.adoc
@@ -11,7 +11,7 @@ The scope of this problem and decision here is explicitly custom services that c
 How cloud instances are provisioned is a different problem with a different decision.
 ====
 
-=== Missing Features In Crossplane 1.6
+=== Missing Features In Crossplane 1.7
 
 [qanda]
 Generate random strings::
@@ -40,7 +40,7 @@ It's currently impossible to apply certain types of patching:
 [NOTE]
 ====
 This list doesn't mean that they should be part of Crossplane, only that they are missing from the PoV of AppCatalog.
-Please see also https://github.com/crossplane/crossplane/issues/2524[#2524], where the community discusses "Custom Compositions".
+Please see also https://github.com/crossplane/crossplane/issues/2524[#2524], where the community discusses 'Custom Compositions'.
 In there the problems with Compositions are summarized pretty well and also AppCat is affected by the limitations of Compositions.
 ====
 
--- a/docs/modules/ROOT/pages/explanations/decisions/crossplane-impl.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/crossplane-impl.adoc
@@ -1,0 +1,90 @@
+= Service Provisioning Implementation
+
+== Problem
+
+Crossplane brings a lot of lifecycle features for XRDs and XRs.
+While developing the https://github.com/vshn/appcat-service-prototype[service prototype] it became apparent that some functionality is missing in Crossplane out-of-the-box.
+
+=== Missing Features In Crossplane 1.6
+
+[qanda]
+Generate random strings::
+See discussions in https://github.com/crossplane/crossplane/issues/1895[#1895] and https://github.com/crossplane/crossplane/pull/2308[#2308].
+Currently the https://github.com/mittwald/kubernetes-secret-generator[Secret generator] is used but this should be regarded as a workaround.
+
+Webhook support for validations and custom business logic::
+See design proposal in https://github.com/crossplane/crossplane/pull/2719[#2719].
+
+Delay reconciliations to specific times when using `CompositionRevisions` feature::
+This is needed to enable maintenance windows.
+
+Cross-resource references within `Composition`::
+This could eliminate a lot of patch duplications.
+See discussions in https://github.com/crossplane/crossplane/issues/1770[#1770] and https://github.com/crossplane/crossplane/issues/2099[#2099].
+
+[NOTE]
+====
+This list doesn't mean that they should be part of Crossplane, only that they are missing from the PoV of AppCatalog.
+====
+
+=== Deployment Requirements
+
+There are several additional resources required along the "usual deployment resources".
+
+[qanda]
+Base service runtime::
+* Helm release
+* Crossplane resources for cloud instances
+
+Backup::
+* S3 Buckets (through Crossplane)
+* K8up `Schedule`
+
+Service Monitoring::
+* Prometheus `ServiceMonitor`
+* Prometheus `PrometheusRule`
+
+Networking::
+* `NetworkPolicy` to allow traffic to the service.
+
+RBAC::
+* `RoleBinding` to grant read access for requesting customer to instance logs.
+
+Customer association::
+Each instance needs to be associated to the requesting customer for billing the services.
+This might require additional business logic.
+
+== Proposals
+
+=== Proposal 1: Umbrella Charts
+
+The primary deployment strategy will be a Helm chart that contains most, if not all deployment artifacts.
+This includes cloud instances provisioned by Crossplane itself.
+
+Additional logic is implemented through custom webhooks and controllers.
+
+.Service provisioning with an umbrella Helm chart
+image::appcat_proposal_1.drawio.svg[]
+
+The "umbrella" chart contain all the feature gates and dependent charts required to provision services on the supported cloud infrastructure.
+The chart takes input parameters such as
+- cloud provider
+- major version
+- backup enabled
+- SLA (alerting enabled)
+
+and then deploys the resources as required.
+
+The chart is also used when cloud instances are provisioned.
+For example, there may be additional Prometheus exporters needed in case the cloud instance doesn't provide Prometheus metrics endpoints.
+
+In addition to the composition, there are webhooks to perform validations and assignment to customer.
+Also, a dedicated version update controller delays changing of `CompositionRevisions` to the newest version until a specific time is met.
+
+Advantages::
+- Umbrella chart is able to deploy arbitrary resources
+
+Disadvantages::
+- Separate version controller required for automated updates
+- Webhook controller required
+- Logic and feature gates in umbrella chart is limited by Helm's templating engine
--- a/docs/modules/ROOT/pages/explanations/decisions/crossplane-impl.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/crossplane-impl.adoc
@@ -152,3 +152,23 @@ Disadvantages::
 - Loose coupling of the resources.
 - A lot of repetition in the compositions.
 - Impact of versioning mechanics of a provider is currently unknown.
+
+=== Proposal 5: Dedicated Provider With Side Chart
+
+This proposal combines the ideas of proposal 3 with proposal 2.
+A Crossplane provider uses code to provision the service, whereas the additional resources are deployed using a Helm chart.
+
+.Service provisioning with a dedicated Crossplane provider and side chart
+image::appcat_proposal_5.drawio.svg[]
+
+Advantages::
+- Flexible deployment using code and Kubernetes API.
+- Webhook controller can be built-in.
+- Built-in version update controller.
+- Generating the resources is unit testable.
+- Common resources are sharable between services using the chart.
+
+Disadvantages::
+- Loose coupling between service and additional resources.
+- Impact of versioning mechanics of a provider is currently unknown.
+- Logic and feature gates in side chart is limited by Helm's templating engine.
--- a/docs/modules/ROOT/pages/explanations/decisions/crossplane-impl.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/crossplane-impl.adoc
@@ -29,7 +29,7 @@ This list doesn't mean that they should be part of Crossplane, only that they ar
 
 === Deployment Requirements
 
-There are several additional resources required along the "usual deployment resources".
+There are several additional resources required along the "usual" deployment resources.
 
 [qanda]
 Base service runtime::
@@ -114,3 +114,22 @@ Disadvantages::
 - Webhook controller required (possibly per service).
 - Logic and feature gates in side chart is limited by Helm's templating engine.
 - Side chart and service definition are loosely coupled.
+
+=== Proposal 3: Dedicated Provider
+
+This proposal uses a dedicated Crossplane provider to deploy and configure all the resources that the service needs.
+A Crossplane provider uses code thus provides the most flexibility.
+
+.Service provisioning with a dedicated Crossplane provider
+image::appcat_proposal_3.drawio.svg[]
+
+Advantages::
+- Flexible deployment using code and Kubernetes API.
+- Webhook controller can be built-in.
+- Built-in version update controller.
+- Strong coupling of the resources.
+- Generating the resources is unit testable.
+
+Disadvantages::
+- There may be a lot of code repetition between services that deploy the same set of common resources.
+- Impact of versioning mechanics of a provider is currently unknown.
--- a/docs/modules/ROOT/pages/explanations/decisions/crossplane-impl.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/crossplane-impl.adoc
@@ -172,3 +172,14 @@ Disadvantages::
 - Loose coupling between service and additional resources.
 - Impact of versioning mechanics of a provider is currently unknown.
 - Logic and feature gates in side chart is limited by Helm's templating engine.
+
+=== Proposal 6: Free Choice
+
+This proposal does not impose a certain strategy how services are to be provisioned.
+Each service can choose how to best provision the required resources in a `Composition`.
+
+Advantages::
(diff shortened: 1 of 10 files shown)


ccremer:
Added a proposal where it's not further defined how services are provisioned

Kidswiss:
I like idea 3. It seems to be the most flexible.

Kidswiss (approved):

glrf:
Just my 2 cents, because I was asked to give input. I did not use Crossplane for quite a while and I don't know much about the current AppCat project state. I think this looks like a solid list of options, I have a few inputs and questions:

* Especially in the case of the dedicated provider. Is the idea to provision Crossplane Helm release resources or would you handle helm templating in the provider?
* One option you skipped (probably for good reason) is option 1, but the Umbrella chart is just a large composition (AFAIK compositions can also deploy arbitrary resources right?) This is probably a horrible idea as we will end up with complex and unmaintainable compositions.


With option 1 and 2 you will probably be able to get something running quickly, but will end up in YAML hell eventually (reminds me of kyverno). With option 3 you reinvent some of the features Crossplane compositions can provide. With option 4 you might end up with rather large and complex compositions.

IMO options 3 and 4 sound like the safest approaches. My gut feeling would be to go with option 4 as it seems to be more in line with Crossplanes philosophy.

docs/modules/ROOT/pages/explanations/decisions/crossplane-impl.adoc
```diff
@@ -0,0 +1,185 @@
...
+
+This proposal uses a dedicated Crossplane provider to deploy and configure all the resources that the service needs.
+A Crossplane provider uses code thus provides the most flexibility.
+
+.Service provisioning with a dedicated Crossplane provider
+image::appcat_proposal_3.drawio.svg[]
+
+Advantages::
+- Flexible deployment using code and Kubernetes API.
+- Webhook controller can be built-in.
```
Kidswiss, line 128: I'm not sure about that one. The providers I've seen so far are only short-lived and are spawned on-demand if changes need to be applied. Maybe it's possible to have always-on providers? If not we'd probably need a seperate controller for still.

docs/modules/ROOT/pages/explanations/decisions/crossplane-impl.adoc
```diff
@@ -0,0 +1,185 @@
...
+- Built-in version update controller.
```
Kidswiss, line 129: Same here.

docs/modules/ROOT/pages/explanations/decisions/crossplane-impl.adoc
tobru: Webhooks are coming in 1.7 next week: https://github.com/crossplane/crossplane/pull/2919

ccremer:
Thanks for your input :+1: 

>  Especially in the case of the dedicated provider. Is the idea to provision Crossplane Helm release resources or would you handle helm templating in the provider?
> 
My naive idea would be to deploy the Crossplane Helm release. The provider would only set the required parameters (which might need dynamic computing based on the instance's spec).

>   One option you skipped (probably for good reason) is option 1, but the Umbrella chart is just a large composition (AFAIK compositions can also deploy arbitrary resources right?) This is probably a horrible idea as we will end up with complex and unmaintainable compositions.
> 
With the umbrella chart I actually meant a Helm chart and not just a large composition. Helm charts have additional functions that Crossplane lacks (e.g. we can't disable certain resources from being deployed or do loops). So yeah, instead of deploying resources via composition, they are deployed in a chart (or multiple charts), to avoid complex compositions. Although, that complexity is outsourced to the chart instead but I believe this to be a bit more maintainable and testable.
> 
> With option 1 and 2 you will probably be able to get something running quickly, but will end up in YAML hell eventually (reminds me of kyverno). With option 3 you reinvent some of the features Crossplane compositions can provide. With option 4 you might end up with rather large and complex compositions.
> 
The YAML hell and Kyverno is exactly the reason why I also prefer option 3 :)
True, but Compositions are insufficient for dynamic variables.

> IMO options 3 and 4 sound like the safest approaches. My gut feeling would be to go with option 4 as it seems to be more in line with Crossplanes philosophy.

Sub providers might be more in line with philosophy, but the composition will become more complex again (what I meant with "loose coupling"). Especially since Cross-patching references in Compositions is impossible. For example, I can't use a generated name or a password of a resource and reference this name in another, a lot would have to be engineered based on assumptions that values of another resource is predictable. Cases like this would be to deploy an S3 bucket with a generative name that is used in the service itself again. Or maybe the namespace name needs to be set for every provider spec since each sub provider essentially has no knowledge of the other.

docs/modules/ROOT/pages/explanations/decisions/crossplane-impl.adoc
ccremer: > The providers I've seen so far are only short-lived and are spawned on-demand if changes need to be applied.

What do you mean by this? The Helm provider for example is nothing but an operator that is reconciling Helm releases.

And with "Webhook controller can be built-in." I meant that it's easy to develop/generate a webhook controller that lives in the same code, but maybe runs in a separate pod just like any other Kubernetes Operator using webhooks. Do you feel I should be more specific about this?

glrf:
> Sub providers might be more in line with philosophy, but the composition will become more complex again (what I meant with "loose coupling"). Especially since Cross-patching references in Compositions is impossible. For example, I can't use a generated name or a password of a resource and reference this name in another, a lot would have to be engineered based on assumptions that values of another resource is predictable. Cases like this would be to deploy an S3 bucket with a generative name that is used in the service itself again. Or maybe the namespace name needs to be set for every provider spec since each sub provider essentially has no knowledge of the other.

That's a valid point, I don't have much experience writing complex compositions. I definitely see the advantage of using a general purpose language and the flexibility it brings. We just need to be aware that this will lead to extra engineering effort.

Also then why do we need the initial composition anyway? As far as I can tell it now doesn't do much anymore.

ccremer:
> Also then why do we need the initial composition anyway? As far as I can tell it now doesn't do much anymore.

That is a good question I don't really have a good answer for. I guess Crossplane/Composition brings a bunch of lifecycle boilerplate like status conditions, events, cleanups. But then again it introduces more complexity and a dependency where I'm not sure if that's really worth...

Crossplane might be useful to deploy cloud instances like RDS or Terraform using Terrajet etc, but we might as well build a standalone Operator.
Such an Operator wouldn't be doing much except deploying opinionated resources.

I have to think about it.

docs/modules/ROOT/pages/explanations/decisions/crossplane-impl.adoc
Kidswiss: > Do you feel I should be more specific about this?

Yes please, I think that's the source of my confusion here.

Kidswiss:
It feels like some mix of Crossplane and custom provider/operator seem the best fit for our needs.

Crossplane can help with Cloud (S3, terraform, etc.) and probably helm management.
Our own provider/operator brings a lot more flexibility for the business logic to the table.

So I see two options:
* We use Crossplane with a single provider: All CRDs and lifecycle are handled within Crossplane, the actual logic resides inside the provider. The provider then creates various other k8s resources, including further Crossplane resources, etc.
* We use an operator. The operator basically handles the lifecycles, CRDs and deploying of further resources. Those resources can also include Crossplane resources.

While writing down those thoughts, it started to sound like implementing our own operator would re-invent Crossplane to some extent... Probably not what we want to do?

My idea would be to go for my first option. We can use Crossplane to define the CRDs and use all other Crossplane providers. But instead of relying on Crossplane to do the composition, we use the provider. It handles all the business logic of creating all the other resources needed. With that, we should be able to keep the compositions in Crossplane relatively simple, as they'd only need to pass the specs to the provider.

I haven't used Crossplane that extensively as @ccremer, so not sure if my idea is feasible.

ccremer:
Thanks for the offline discussion @Kidswiss and @glrf 

Summary of our call:

The discussion was about our two favourite proposals 3 and 7 (provider vs operator)

* The Crossplane community is growing, however AppCat services are likely to become too opinionated to contribute back to the Crossplane community. So neither we nor the community will benefit if the provider is not really sharable.
* We expect differences how the service architectures look like (Helm-based installations vs Cloud instances). Currently we don't exactly know which extra effort would be needed to provision an AWS RDS cloud instance with the AppCat extras for example. If it turns out to be completely fundamentally different, we would consider different Crossplane providers to be slightly superior than dedicated operators. For example, a customer might choose MariaDB compositions using the `provider-mariadb-on-aws` or `provider-mariadb-on-helm` providers.

Next step: Find out what the "AppCat Extras" are for an AWS RDS instance.

tobru:
Thanks for the list of options. I studied it in detail and I have some thoughts and questions.

> AppCat services are likely to become too opinionated to contribute back to the Crossplane community.

Why that? I understand Crossplane to be exactly what we need for AppCat. Where would AppCat deviate from what the idea behind Crossplane is?

From what I learned about Crossplane and from all the options presented "Proposal 4: Sub Providers" feels to be most matching the concepts. All other options do not really match Crossplane's idea.

The main question I asked myself before studying the options: What exactly do we need to do for an AppCat service? It's about generating Kubernetes resources based on a defined input (XRD). The current list of proposal mainly uses Helm Charts or Custom Providers for that, but there also is [provider-kubernetes](https://github.com/crossplane-contrib/provider-kubernetes) which is able to generate arbitrary Kubernetes resources, which could create the additional AppCat objects required, besides what the Service Helm Chart provides.

What I miss is a link to Commodore Components. Compositions could be rendered by Commodore Components to not have to duplicate content, let Jsonnet do that.

Regarding validation webhooks: We have Kyverno which is able to validate XRs, we don't have to write custom webhooks in Crossplane for that (see e.g. [Enforcing Kubernetes Policies On Infra And Apps By Combining Kyverno And Crossplane](https://www.youtube.com/watch?v=sxASDgKu0dQ)).

For the automated maintenance I can see a need for a custom controller which handles setting `CompositionRevisions` based on rules.

All in all I currently can't see where Crossplane Compositions can't handle our use-case.

ccremer:
I did some research on Google SQL and AWS RDS and how it could look like in AppCat context.
I came to the conclusion that it actually makes little to no sense to resell cloud instances under AppCat. This is how I see things:

* Both GCP and AWS have built-in performance insights and logs that are hard for us to get. And even if we do get metrics, I'm pretty sure our dashboards won't be half as good as AWS/GCP already provide. The same is for logs: The only way to access logs is by using their API, we would need to write some logging proxy by using their SDK or REST endpoints.
* We have almost no influence on backups. At least for RDS, we can't really fulfill the requirement to export backups in a neutral format. Apparently RDS uses a custom format that is only readable with AWS tools.
* Both AWS and GCP offer alerting rules, so it should be easy to configure notifications for customers in case something happens with their instance.

I see 2 possible options to go forward:
* We offer/resell cloud instances under AppCat, but without logs and metrics and Prometheus-based alerting. Since we would artificially limit what GSQL and RDS can offer, I see no reason to re-invent the wheel that GSQL and RDS already could offer. 
* We don't offer/resell cloud instances under AppCat, but let the customers provision instances with their own account. We could consider installing Crossplane providers, but they'd use their account in order to get the full experience (insights, logs, backups etc)

I see option 2. The complexity to bring logs and metrics from cloud instances into AppCat is not worth the effort IMO. The chances are high that we install helm-based database services on clusters even where cloud instances are possible, so customers "that just want a database" can use this. Customers that actively want cloud instances should get the full features, not limited or reinvented by AppCat.

tobru:
It makes perfectly sense to resell cloud services under the AppCat umbrella. The main reason is that even for a cloud service to properly work in a "perfect" manner, a lot of knowhow and configuration is needed. An example: An AppCat user orders a Managed Database, based on AWS RDS: Many knobs have to be turned so that this instance adheres to our quality requirements, this isn't done automatically. That's where the Crossplane Composition comes into play and provisions the instance and sets all needed parameters to our own defaults, so it matches the quality requirements of a service from the AppCat.

I'm all in to use the cloud providers possibilities for logging and monitoring, no-one wants us to reinvent the wheels. What we would have to do in our automation is to configure monitoring, so that it sends alerts to a well-defined destination, write runbooks, configure logging so that it can be accessed by the customer and so on.

ccremer:
> > AppCat services are likely to become too opinionated to contribute back to the Crossplane community.
> 
> Why that? I understand Crossplane to be exactly what we need for AppCat. Where would AppCat deviate from what the idea behind Crossplane is?

So what IS the idea behind Crossplane? Let's take Redis as an example. Our Swiss-based cloud providers don't offer cloud instances so we have to engineer the whole stack, including logs, metrics and backups. If we choose K8up as the backup solution it's already opinionated. Meaning the community would have to use, understand and operate K8up in addition to Redis. That's why I think it's too opinionated so that other users would not make use of it.
 
> From what I learned about Crossplane and from all the options presented "Proposal 4: Sub Providers" feels to be most matching the concepts. All other options do not really match Crossplane's idea.

Why do we have to match Crossplane's idea in the first place? The customer couldn't care less about Crossplane since all they'd use is a CRD (either generated by Crossplane or from a custom Operator) as the interface to provision services. Provisioning services using Crossplane is "merely" an implementation detail that can result in high complexity.

> The main question I asked myself before studying the options: What exactly do we need to do for an AppCat service? It's about generating Kubernetes resources based on a defined input (XRD). The current list of proposal mainly uses Helm Charts or Custom Providers for that, but there also is [provider-kubernetes](https://github.com/crossplane-contrib/provider-kubernetes) which is able to generate arbitrary Kubernetes resources, which could create the additional AppCat objects required, besides what the Service Helm Chart provides. See e.g. [crossplane-contrib/provider-kubernetes#13](https://github.com/crossplane-contrib/provider-kubernetes/issues/13) for what the latest release of provider-kubernetes is capable to do: "Patch resource with values by resolving from its dependent resources."

provider-kubernetes doesn't really solve the problem where values have to be computed/modified from the instance spec. It just moves the problem one abstraction deeper. Besides, the objects seem to be cluster-scoped and even more YAML: https://github.com/crossplane-contrib/provider-kubernetes/blob/main/examples/object/references/patches-from-multiple-resources.yaml

> What I miss is a link to Commodore Components. Compositions could be rendered by Commodore Components to not have to duplicate content, let Jsonnet do that.

While I generally agree, it still doesn't fully solve the computed values problem. And even if the composition is rendered using jsonnet, there are chances that they result in hundreds of lines of YAML (due to the patching rules to make it somewhat work), which certainly doesn't help in troubleshooting or developing and testing services. Testing and maintaining such compositions is very hard.

> Regarding validation webhooks: We have Kyverno which is able to validate XRs, we don't have to write custom webhooks in Crossplane for that (see e.g. [Enforcing Kubernetes Policies On Infra And Apps By Combining Kyverno And Crossplane](https://www.youtube.com/watch?v=sxASDgKu0dQ)).

Let's use Kyverno only as the absolute last resort. The whole Tarazed team regrets the decision to use Kyverno for APPUiO Cloud to implement business logic with Kyverno policies for multiple reasons, so let's no repeat the same mistake. I am very strongly for custom webhooks where custom logic is needed (and it most probably is needed where we have to assign ownership to service instances for billing)

> All in all I currently can't see where Crossplane Compositions can't handle our use-case.

They can handle our use-cases in some war or another, the question is more "is it the *right* thing" to handle our use-cases?

ccremer:
> It makes perfectly sense to resell cloud services under the AppCat umbrella. The main reason is that even for a cloud service to properly work in a "perfect" manner, a lot of knowhow and configuration is needed. An example: An AppCat user orders a Managed Database, based on AWS RDS: Many knobs have to be turned so that this instance adheres to our quality requirements, this isn't done automatically. That's where the Crossplane Composition comes into play and provisions the instance and sets all needed parameters to our own defaults, so it matches the quality requirements of a service from the AppCat.

I do see the reason to provide some defaults and I didn't exclude the possibility that we can setup compositions. But I think the instance should then be owned by the customer so it can be access with their account. However, our management doesn't go much further here. I don't see how we should come up with our own set of backup, maintenance and update policies for example. Alerting is a different story, see below.

> I'm all in to use the cloud providers possibilities for logging and monitoring, no-one wants us to reinvent the wheels. What we would have to do in our automation is to configure monitoring, so that it sends alerts to a well-defined destination, write runbooks, configure logging so that it can be accessed by the customer and so on.

> configure monitoring, so that it sends alerts to a well-defined destination

That is possible for GSQL, but only using provider-jet-gcp using Terraform. However, I could not figure out how we can do the same for AWS RDS, neither provider-aws nor provider-jet-aws offer anything that remotely hints at alert rules. So we would have to use AWS CloudWatch for this, for which I haven't seen anything that Crossplane offers,  we'll have to engineer using the AWS SDK or maybe some other solutions exists. 
My point is: Some cloud providers are easier to integrate, other not so, maybe even impossible, it'll be hard to find common denominators.

> configure logging so that it can be accessed by the customer

This could end up in a big yak if they don't have access to the reselling account. Can we at least consider the possibility that the customers would have to bring their own accounts if they want cloud instances (not using a shared account)? That would solve a lot of issues and they can view logs and metrics on their own without having us to engineer any additional Grafana dashboards or so. I'm assuming that customers shouldn't have access to cloud console using the shared, cluster-wide account unless it's a private cluster.

ccremer:
From this discussion, may I propose the following:

- Cloud services and non-cloud-services (e.g. Helm) are fundamentally different. We can't and shouldn't treat them equally where we can "just abstract the common denominators", and this goes up to the requirements. Meaning the CRDs required to provision cloud and non-cloud services have to be completely different, they can't share a similar spec and using either of the two architecture is a conscious decision.
- Once we treat the architecture differently, I would propose to split the decision/PR for each type and decide separately how to implement it. Since these types have barely anything in common.
- I would like to engineer a separate prototype for a cloud instance. Most likely we need to use Terrajet for this since provider-gcp and provider-aws are likely insufficient for our use cases and I don't have any clue what it means to use any of the (terrajet) providers. It would also give us a better answer for the logging, metrics and alert problems.

tobru:
While reading the answer I feel that there are a lot of assumptions about things we have not worked out yet, e.g. which account to use for cloud services. That's not defined as of now and still up to figure out. The current focus lies on locally provisioned services, cloud services come later.

The reason why I keep poking on Crossplane is that I would like to see a common framework across the services we offer in AppCat and I don't want to reinvent the wheel and also not write our own framework from scratch. That's also the reason why I proposed Kyverno as the governor and policy engine, let's not reinvent our own policy framework.

ccremer:
Thanks for the offline discussion today @Kidswiss @tobru 

Summary:
- We agree and acknowledge that Crossplane `Compositions` are limited and these alone can't achieve what we have in mind with AppCat
- We agree and acknowledge that deploying resources via umbrella or side Helm charts is not flexible or testable enough.
- We agree that writing code is currently the most promising way forward

We currently have 3 favorites: Proposal 3, 7 and (new) 8

---

I have put a bit of thought into proposals 3 and 8 and compared them.
I would formulate the decision as following:

#### Decision
Proposal 3 (dedicated provider)

#### Rationale
Crossplane `Composition` API is limited and it alone can't solve the problem mentioned at the top.
While umbrella and side charts sound easy at first, they lack flexibility and testability compared to code running in some form of controller.
Proposal 4 is risking a too lose coupling between the resources.
An totally undefined foundation as in proposal 6 makes puts operational aspects into silos, requiring a learning curve for each individual service.
Through method of elimination, it is evident that proposal 3, 7 and 8 are remaining.

Compared to proposal 7, proposal 3 is favored for the following reasons:
- It's expected to need less engineering effort in resource lifecycle management at the cost of higher complexity in operations.
- Assuming the future decision of provisioning cloud instances also with Crossplane, they would be aligned and comparable.
  This would mean a steeper learning curve to understand the Crossplane stack in both development and operations, but once understood should be more efficient.
  However this can be considered a risk if the decision is not Crossplane, in which case it may be sensible to transform the provider into its own operator, since the overhead complexity of Crossplane is not exactly required or portentially not even justified.
  This would effectively become proposal 7.

Compared to proposal 8, proposal 3 is favored for the following reasons:
- While some experience could be gained by developing an AppCat service prototype, the actual details, dependencies and relations between common resource and service-specific resources are currently still rather uncertain.
  For this reason, it makes sense to start the first productive service with strong coupling by having the common artifacts in the same provider and internal code.
- Once the dependencies and relations could be made out, it may be sensible to refactor and extract the common artifacts into their own provider.
  This would effectively become proposal 8.

---

What do you think? Is this agreeable or would you like another offline discussion?

tobru:
Thank you very much @ccremer for the work in this decision.

> What do you think? Is this agreeable or would you like another offline discussion?

I'm absolutely fine with the proposed decision and the rationale behind :+1:

tobru (approved):

PR #30, opened 2022-04-21
Decision for converged service implementation: Use dedicated Operator

ccremer:
## Summary

This PR revisits the decision regarding converged service implementation.
Internal discussions with the AppCat engineers revealed that the added complexity of integrating service provisioning into Crossplane doesn't add much value. A dedicated Operator is more simple to understand, maintain and operate in production.

This PR replaces the decision to use Proposal 7 with Proposal 8, with the option that Crossplane may be added later on top of it, if so desired.
Whether Crossplane is used or not has no effect for the end user and thus it should be considered an "implementation detail".

Reverts/Adjusts #19 

## Checklist


ZuGao:
LGTM

tobru:
> Whether Crossplane is used or not has no effect for the end user and thus it should be considered an "implementation detail".

While I agree that this is true for the "backend implementation", this isn't true for the "frontend", the resources which the user is interacting with. When using Crossplane CRs, they follow the Crossplane Resource Model (XRM) and behave the same, independent of the backend implementation. I suggest adapting "Proposal 7" so that Crossplane Compositions are used, not just as an option.

> Going for the most simplest and flexible option first (a dedicated Operator) still allows platform engineers to build Crossplane Compositions that make use of the service operator.

Who is "platform engineers"? IMHO that's also part of AppCat to provide the Crossplane Compositions.

Some other thoughts:
* Somehow while thinking about this topic I was reminded of Carpenter: https://syn.tools/syn/SDDs/0023-managed-services-controller.html. We tried to implement something back then already. Just as a reminder.
* Going through all the proposals again, I'd see another proposal, a slight adaption to number 8, replacing provider with operator:
  * Application aware Operator to deploy the application itself, instead of Helm Charts, if there is a usable and well-maintained Operator available. Otherwise, use Helm Charts.
  * Separate "AppCat" Operator which does all the "AppCat" things (K8up, S3 Bucket, Monitoring, ...), somehow what we once thought to do with Carpenter
  * Composition which composes these resources.

What I want to try to circumvent is that we reinvent the wheels with yet another application Operator. So if there already is a well-maintained and good working Redis Operator available, we should use it, instead of writing our own Operator.

ccremer:
> When using Crossplane CRs, they follow the Crossplane Resource Model (XRM) and behave the same, independent of the backend implementation.

This abstraction was deemed not worthy to follow up for the following reasons:
1. Even for seasoned Kubernetes engineers, the Crossplane resource model is very hard to understand, maintain and troubleshoot. We'd like to avoid depending on it unless absolutely necessary.
2. We've already established that converged service and cloud service is just not the same and it's very hard to find common grounds in the details. We get the idea that one might desire to build a frontend for it, but when it comes to the details soon one will find differences and needs to adjust for a specific backend implementation. The ability to "move" from one platform to the other without touching the deployment (which includes an AppCat instance) is not yet worth this abstraction. There's barely any customer application migration that doesn't touch the deployment at all.

> I suggest adapting "Proposal 7" so that Crossplane Compositions are used, not just as an option.

We opted to start without Crossplane compositions and leave that option open for later. It should be possible that at some point we can run both at the same time. May I suggest to leave that as an option until we've decided to actually use compositions on top of the operator?

> Somehow while thinking about this topic I was reminded of Carpenter: https://syn.tools/syn/SDDs/0023-managed-services-controller.html. We tried to implement something back then already. Just as a reminder.

No, we never "tried to implement" anything in Carpenter. There's not a single commit in https://github.com/projectsyn/carpenter that remotely hints at implementation besides the boilerplate. It never was more than an SDD.
But yes, although the concept is similar, the difference is that the resources are managed in code and not in Yaml, which allows us to simplify the API spec in the CRD massively and make it consumer-friendly.

> Application aware Operator to deploy the application itself, instead of Helm Charts, if there is a usable and well-maintained Operator available. Otherwise, use Helm Charts. 

The proposals that are based on provider or operator don't mention that they _require_ to deploy helm charts. These are just ideas. Yes, I agree we should look into native operators, however, I'm dead certain that whatever deployment method, we need to "glue" more resources and make them VSHN-opinionated. That's the core purpose of the Operator or Provider along with a maintenance-window/update controller. Although this reminds of Compositions again which have a similar purpose, but as already established, they are not suitable enough for our use cases.

> Separate "AppCat" Operator which does all the "AppCat" things (K8up, S3 Bucket, Monitoring, ...), somehow what we once thought to do with Carpenter

Yes, this separation is possible and likely in an operator, as a code library or via composition. For now it's too early to tell. Let's start with the first service implementation and see what we need for the 2nd or 3rd.

> What I want to try to circumvent is that we reinvent the wheels with yet another application Operator. So if there already is a well-maintained and good working Redis Operator available, we should use it, instead of writing our own Operator.

We have to decide service-by-service what the best option is. We've already had 2 cases where Helm charts upstream made significant breaking changes and we forked them and modified some health probes... Upstream Operators might just do the same as well and do things that don't meet or conflict with our requirements.

In short, every time we use upstream deployments we are the mercy of the upstream maintainers. If we do it ourselves, we have it under control but the maintenance effort is higher. Classic make or buy decision. Sometimes buying is better, sometimes making is better, even if it means reinventing the wheel. Choose your poison? :)

Let's start with the Postgresql operator to get _somewhere_

tobru (approved):

Kidswiss (approved):

PR #31, opened 2022-07-08
Application Catalog Documentation - Requirements - What is it - Why

tobru:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/explanations/what_is.adoc
+++ b/docs/modules/ROOT/pages/explanations/what_is.adoc
@@ -1,0 +1,34 @@
+= What is the Application Catalog
+
+TL;DR::
+* A catalog of services which can be self-service ordered directly from a Kubernetes cluster
+* A framework to build services which are then available in the catalog
+
+[quote, products.docs.vshn.ch]
+____
+The VSHN Application Catalog is a Cloud Native marketplace which offers services from Cloud Service Providers like Amazon AWS, Google Cloud, Microsoft Azure, Aiven.io, Exoscale or cloudscale.ch, as well as managed services from VSHN.
+____
+
+Behind the scenes it's *our framework* to self-service deploy applications on Kubernetes, by simply ordering a new instance directly in the cluster (no external system involved).
+It's a list of applications available in an APPUiO Kubernetes cluster from which the user of the cluster can choose from.
+The list of applications heavily depends on where the cluster runs and it's capabilities.
+
+== The Framework
+
+It enables us to:
+
+* Easily engineer new services for the application catalog, by still having them to adhere to our standards and make them discoverable
+* Have similarity between all services in the application catalog because they all are built using the same framework and technology
+* A standardization of the same service between clusters, because we set our defaults and only expose the parameters we want to allow to be customized
+
+=== A comparison
+
+For a better picture, the framework can be compared to:
+
+Puppet::
+* Base modules allow installing a service and configure (usually) every aspect of it (a lot of parameters are exposed)
+* Profiles use base modules, glue them together and configure them with our own best-practice, to form a new service. Usually only a few parameters are exposed and only these parameters can be influenced.
+
+Project Syn::
+* Helm charts are the base modules. They allow installing and configure the service and expose all possible options.
+* Commodore Components are sticking Helm Charts together, parametrize them with best-practices and only make a handful of options available to the user.
--- a/docs/modules/ROOT/pages/explanations/why_exists.adoc
+++ b/docs/modules/ROOT/pages/explanations/why_exists.adoc
@@ -1,0 +1,56 @@
+= Why does Application Catalog exist?
+
+The application catalog is the successor to the classic VSHN Managed Services on virtual machines, managed by Puppet.
+It's a very similar concept, applied to Kubernetes, but extended with Cloud services integration, which wasn't available with Puppet.
+
+More resources:
+
+* https://products.docs.vshn.ch/products/appcat/index.html#_use_cases[Use Cases] on the product site
+* https://handbook.vshn.ch/vshn_key_deliverables.html[Key Deliverables]
+* https://handbook.vshn.ch/vshn_customer_segments.html#_instances_at_scale_and_saas[Customer Segments] - mainly "Instances at Scale and SaaS"
+
+== For the user
+
+There are different users:
+
+* The VSHNeer who engineers the application hosting environment for a VSHN customer
+* The APPUiO Cloud user who needs a service for their application running on APPUiO Cloud
+* The APPUiO Managed user who cares about the application themselves
+* The developer who needs a service for the application but doesn't want to be concerned about platform topics
+
+=== Consume services for the application
+
+Software running on a Kubernetes cluster usually needs one or many backend services (databases, caches, object storage, and so on).
+Ordering and consuming of such a service must be as simple as possible to not cause friction and the focus can lie on the application.
+
+=== Self-Service with Best-Practices
+
+When a service from the application catalog is needed, it must be available straight away without waiting for someone to provision it.
+Best-practices should already be applied by the platform team, which isn't the developers concern.
+
+== For VSHN
+
+=== Enable self-service for Developers
+
+By enabling the developer to self-service order a service we can focus on providing the service, instead of provisioning of the service.
+
+=== Simplify service brokering
+
+Brokering of external services gets much easier when we have a set of pre-defined services available, rather than reinventing the provisioning and cloud architecture every time again and again.
+
+=== Standard way to provide services
+
+All provided services look the same as they were provided by the same framework.
+
+== For a Cloud Service Provider
+
+A cloud service provider might want to onboard new customers and needs to provide more services.
+
+== For a software vendor
+
+A software vendor might want to provide their software in a Kubernetes cluster and have it as a part of the application catalog.
+
+== Differentiation to Project Syn
+
+Project Syn, namely Commodore Components, are static configuration options for a whole Kubernetes cluster.
+It lacks the self-service aspect completely and hasn't been built for that.
--- a/docs/modules/ROOT/pages/reference/framework-requirements.adoc
+++ b/docs/modules/ROOT/pages/reference/framework-requirements.adoc
@@ -1,0 +1,36 @@
+= Framework Requirements
+
+As discussed on xref:explanations/what_is.adoc["What is it?"], a *framework* for self-service provisioning of services out of a Kubernetes cluste is part of the application catalog.
+
+This page discusses requirements the framework must adhere to:
+
+Reusability::
+Base parts of the framework should be reusable by the Open Source community and VSHNeers.
+Just like Puppet base module or Helm charts are (can be) reusable.
+
+Low barrier to entry - xref:reference/glossary.adoc#_personas[Service Maintainer]::
+Adding a new service to the catalog using the framework should be as easy as possible.
+Required skills:
+Must:::
+* Kubernetes
+* Helm Charts
+* (Crossplane)
+Should:::
+* Commodore Components
+
+Low barrier to entry - xref:reference/glossary.adoc#_personas[Service User]::
+Using a service should be a no-brainer:
+* Simple interface
+* Expressive and approachable documentation
+
+Low maintenance::
+Our focus must be on maintaining the service itself.
+Core parts of the framework should be maintained by a community.
+(Just like Puppet itself is maintained by Puppet Labs - we maintain the modules and run the infrastructure)
+
+More buy than make::
+Try to use what's already available and focus on sticking parts together to be of use for us.
+Only engineer code (e.g. Go) when it's for sticking parts together or a mandatory part is missing.
+
+Be aware of the risks::
+When "buying" a part of the system, do a risk assesment and document it properly.
--- a/docs/modules/ROOT/pages/index.adoc
+++ b/docs/modules/ROOT/pages/index.adoc
@@ -1,14 +1,11 @@
 = VSHN Application Catalog
 
-image::vshn_logo.png[VSHN,256,link=https://www.vshn.ch]
-
 Welcome to the https://www.vshn.ch[VSHN] Application Catalog technical documentation site.
+The product documentation can be found on the https://products.docs.vshn.ch/products/appcat/index.html[VSHN products documentation site].
 
-Here you’ll find all the technical documentation.
-Use the search field at the top of this page, navigate through topics with the tree on the left or see the links below.
+Recommended pages to start with:
 
-The product documentation can be found under https://products.docs.vshn.ch/[products.docs.vshn.ch].
-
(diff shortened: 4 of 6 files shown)


docs/modules/ROOT/pages/explanations/what_is.adoc
```diff
@@ -0,0 +1,34 @@
...
+TL;DR::
+* A catalog of services which can be self-service ordered directly from a Kubernetes cluster
+* A framework to build services which are then available in the catalog
+
+[quote, products.docs.vshn.ch]
+____
+The VSHN Application Catalog is a Cloud Native marketplace which offers services from Cloud Service Providers like Amazon AWS, Google Cloud, Microsoft Azure, Aiven.io, Exoscale or cloudscale.ch, as well as managed services from VSHN.
+____
+
+Behind the scenes it's *our framework* to self-service deploy applications on Kubernetes, by simply ordering a new instance directly in the cluster (no external system involved).
```
Kidswiss, line 12: > no external system involved

That's not true for resold services. Or what exactly is an "external system" in this context?

docs/modules/ROOT/pages/explanations/what_is.adoc
```diff
@@ -0,0 +1,34 @@
...
+It's a list of applications available in an APPUiO Kubernetes cluster from which the user of the cluster can choose from.
+The list of applications heavily depends on where the cluster runs and it's capabilities.
+
+== The Framework
+
+It enables us to:
+
+* Easily engineer new services for the application catalog, by still having them to adhere to our standards and make them discoverable
+* Have similarity between all services in the application catalog because they all are built using the same framework and technology
+* A standardization of the same service between clusters, because we set our defaults and only expose the parameters we want to allow to be customized
```
Kidswiss, line 22:
```suggestion
* A standardization of the same service between clusters, because we set our best practises and only expose the parameters we want to allow to be customized
```

docs/modules/ROOT/pages/reference/service-requirements.adoc
```diff
@@ -0,0 +1,85 @@
+= Service Requirements and Iterations
+
+Implementing a service is iterative work.
+The idea is to release early and release often, until the desired functionality, stability and maturity is reached.
+
+This page gives an idea on the iterative steps.
+Except the initial version and the final version, all steps in-between can be in any order, as it fits best.
+
+Important is that every step already has the desired quality.
```
Kidswiss, line 9: I'm wondering:

What about external services where we may not be able to cover all the steps? Also for converged services on the cluster. What if implementing maturity "Marbas" would be so much effort for a specific service that it may not be worth it to implement? Should we allow some leeway for each service in respect to the requirements?

docs/modules/ROOT/pages/explanations/what_is.adoc
tobru: I rephrased this part as it doesn't make sense when I've read it again.

docs/modules/ROOT/pages/reference/service-requirements.adoc
```diff
@@ -0,0 +1,85 @@
+= Service Requirements and Iterations
+
+Implementing a service is iterative work.
+The idea is to release early and release often, until the desired functionality, stability and maturity is reached.
```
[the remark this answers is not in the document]
tobru: Rephrased and added more text to clarify that. What do you think about the current state now?

docs/modules/ROOT/pages/reference/service-requirements.adoc
tobru: I've added a few sentences to incorporate this concern. What do you think about it now?

docs/modules/ROOT/pages/reference/service-requirements.adoc
```diff
@@ -0,0 +1,88 @@
+= Service Requirements and Iterations
+
+Implementing and maintaining a service is iterative work.
+The idea is to release early and release often, without having the urge to have all features available right from the beginning.
+
+This page gives an idea on the iterative steps.
+Except the initial version and the feature complete version, all steps in-between can be in any order, as it fits best.
+With "feature complete" it's not meant that the service doesn't receive new functionality or doesn't get maintained anymore, we're talking about the default scope outlined on this page.
```
ccremer, line 8: I have a question regarding the development of the framework itself.
It's my understanding that the framework should help the services covering the iterations outlined below. By extension it means the framework must be capable of supporting these iterations, correct?
Do we know develop or choose a technology that can cover all of this from the beginning, or will there be a complex mix of various technology stack to achieve these requirements with the framework as they come up?

docs/modules/ROOT/pages/reference/service-requirements.adoc
Kidswiss: I would call it `desired quality` but maybe focus on the fact, that a step doesn't imply the implementation of another step. Maybe something like this:

```suggestion
Important is that every step is independent of each other. A service that implements `Abigor` doesn't necessarily already implement `Melchom `.
```

docs/modules/ROOT/pages/explanations/what_is.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+For a better picture, the framework can be compared to:
+
+Puppet::
+* Base modules allow installing a service and configure (usually) every aspect of it (a lot of parameters are exposed)
+* Profiles use base modules, glue them together and configure them with our own best-practice, to form a new service. Usually only a few parameters are exposed and only these parameters can be influenced.
+
+Project Syn::
+* Helm charts are the base modules. They allow installing and configure the service and expose all possible options.
+* Commodore Components are sticking Helm Charts together, parametrize them with best-practices and only make a handful of options available to the user.
```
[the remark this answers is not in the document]
tobru: I see, I have to check that in more detail. And I also see a need to clarify the personas in more detail here...

docs/modules/ROOT/pages/explanations/why_exists.adoc
```diff
@@ -0,0 +1,56 @@
...
+
+By enabling the developer to self-service order a service we can focus on providing the service, instead of provisioning of the service.
+
+=== Simplify service brokering
+
+Brokering of external services gets much easier when we have a set of pre-defined services available, rather than reinventing the provisioning and cloud architecture every time again and again.
+
+=== Standard way to provide services
+
+All provided services look the same as they were provided by the same framework.
```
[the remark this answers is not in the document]
tobru: It clearly isn't clear what I want to say here, I need to come up with something better and define the perspective.

docs/modules/ROOT/pages/reference/service-requirements.adoc
```diff
@@ -0,0 +1,88 @@
...
+
+As a user of the service I can:
+
+* Enable alerting on pre-defined SLIs to VSHN
+* See SLO reports
+* Access operations runbooks
+
+See also:
+
+* xref:reference/quality-requirements/usability/actionable-alerts.adoc[Actionable Alerts]
```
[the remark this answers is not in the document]
tobru: Good point, I'll go over them again.

docs/modules/ROOT/pages/reference/service-requirements.adoc
tobru: Not sure yet, I have to think about that point...

docs/modules/ROOT/pages/reference/framework-requirements.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+Low barrier to entry - xref:reference/glossary.adoc#_personas[Service User]::
+Using a service should be a no-brainer:
+* Simple interface
+* Expressive and approachable documentation
+
+Low maintenance::
+Our focus must be on maintaining the service itself.
+Core parts of the framework should be maintained by a community.
+(Just like Puppet itself is maintained by Puppet Labs - we maintain the modules and run the infrastructure)
```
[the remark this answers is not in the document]
tobru: Depends on the point-of-view. Maintaining Puppet itself might be very complex (outsourced to Puppet Labs), but maintaining Puppet modules is fairly easy (as long as you understand how to engineer Puppet modules). Understanding of technology can be learned. My stake is to lower the amount of technology we use over time to stick to as few as needed.

Looking at Project Syn: Only a few VSHNeers are able to maintain Commodore and Lieutenant, but a lot of VSHNeers can write and maintain Commodore Components. The same should apply to the AppCat Framework: It's OK if just a few VSHNeers can maintain the Framework, but a lot of VSHNeers should be able to engineer and maintain AppCat services.

I'll see how I can introduce that into the requirements.

docs/modules/ROOT/pages/explanations/what_is.adoc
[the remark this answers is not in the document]
tobru: After talking to some Project Syn maintainers, I figured out that my description wasn't 100% correct. There are Commodore Components which are setting default values, but still expose the full Helm values as a fallback. Amended the description accordingly.

docs/modules/ROOT/pages/explanations/why_exists.adoc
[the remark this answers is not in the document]
tobru: I reworded this part to be more precise.

docs/modules/ROOT/pages/reference/framework-requirements.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+More buy than make::
+Try to use what's already available and focus on sticking parts together to be of use for us.
+Only engineer code (for example Go or Python) when it's for sticking parts together or a mandatory part is missing.
+
+Be aware of the risks::
+When "buying" a part of the system, do a risk assesment and document it properly.
```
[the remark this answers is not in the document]
tobru: Added this aspect, thanks for the input.

docs/modules/ROOT/pages/reference/framework-requirements.adoc
```diff
@@ -0,0 +1,46 @@
+= Framework Requirements
+
+As discussed on xref:explanations/what_is.adoc["What is it?"], a *framework* for self-service provisioning of services out of a Kubernetes cluster is a core part of the application catalog.
+
+This page discusses requirements the framework must adhere to:
+
+Reusability::
+Base parts of the framework should be reusable by the Open Source community and VSHNeers.
```
ccremer, line 8: From a business perspective: Why does the framework need to be reusable by the Open source community? The open source community is unlikely going to help the VSHN-opinionated framework. (I'm not saying it couldn't be Open source, but making it _reusable_ adds more complexity in various places). Similar to APPUiO Cloud, where we thought we apply the resellable aspects, but it's not widely used out in the open and still very VSHN-opinionated.

docs/modules/ROOT/pages/reference/framework-requirements.adoc
```diff
@@ -0,0 +1,46 @@
...
+* Kubernetes
+* Helm, Helm Charts
+* Prometheus, PromQL
+* (Crossplane)
+Should:::
+* Commodore Components
+
+Low barrier to entry - xref:reference/glossary.adoc#_personas[Service User]::
+Using a service should be a no-brainer:
+* Simple and discoverable interface
```
ccremer, line 25: What does "discoverable" mean exactly?

docs/modules/ROOT/pages/reference/framework-requirements.adoc
```diff
@@ -0,0 +1,46 @@
...
+Be aware of the risks::
+When "buying" a part of the system, do a risk assessment and document it properly.
+
+Testing facilities::
+Automated testing is important for the quality of a service.
+The framework must provide facilities for a service maintainer to be able to specify tests for a service.
+
+Identity::
+Services which are making use of the framework have a unique identity.
+To achieve that, the framework must help with that.
```
ccremer, line 46: I do not understand what is meant by this. In another paragraph it's required that the services have the same look, yet here we're talking about unique identity? What is meant with "identity"?

docs/modules/ROOT/pages/reference/framework-requirements.adoc
tobru: With that part I meant when we use Crossplane that we write a "base part" provider-cloudscale which is reusable. How should I reformulate that part?

Again compare with Puppet: When I needed to install and configure Apache I wrote a reusable base module for Apache and an opinionated non-reusable profile.

docs/modules/ROOT/pages/reference/framework-requirements.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+Low barrier to entry - xref:reference/glossary.adoc#_personas[Service User]::
+Using a service should be a no-brainer:
+* Simple and discoverable interface
+* Expressive and approachable documentation
+
+Low maintenance::
+Our focus must be on maintaining the service itself.
+Core parts of the framework should be maintained by a community whenever possible or by the framework maintainer.
+(Just like Puppet itself is maintained by Puppet Labs - we maintain the modules and run the infrastructure)
```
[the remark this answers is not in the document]
ccremer: My question had "completionness" and reasoning in mind. I'm fine leaving it as is.

docs/modules/ROOT/pages/reference/framework-requirements.adoc
tobru: I reshuffled the page a bit which should make this requirement more clear.

docs/modules/ROOT/pages/reference/framework-requirements.adoc
tobru: For example its self-documenting so that something like "kubectl explain" works.

docs/modules/ROOT/pages/reference/framework-requirements.adoc
ccremer: Thanks. Though I still don't understand the "identity" bit. Is this some sort of unique name to avoid duplication?

tobru: With identity, I mean a way to identify a service to be part of appcat.

Examples:
* All AppCat services share the same API group (e.g. `appcat.vshn.io/vX`)
* All AppCat services have a common XRD structure, so that common parameters between services are identical (e.g. `spec.sla`)

ccremer: Thanks for the explanation. I would never have guessed that the API group and consistency is meant with this :see_no_evil:  I guess this is confusing for others too. Can we maybe rephrase that to `Services which are making use of the framework share a common API scheme` or something along this? It's clearer than mentioning "identity" without explaining it somewhere.

Kidswiss: I agree. We don't have to specify it in detail yet, but the intention should be made clear.

Kidswiss (changes requested):
See discussion about the API identity.

ccremer (approved):

Kidswiss (approved):

PR #32, opened 2022-07-27
Add PostgreSQL Decision Page

Kidswiss:
## Summary

Add a short page to make a decision about how to proceed with the new approach.

## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
@@ -2,18 +2,23 @@
 
 == Problem
 
-Creating our own operator to deploy and manage PostgreSQL proved to be a gargantuan effort. In order to keep in line with the newly specified https://kb.vshn.ch/app-catalog/reference/framework-requirements.html#_for_service_maintainer["more buy than make"] approach, we're evaluating the viability of using existing operators for the use of converged services.
+Creating our own operator to deploy and manage PostgreSQL proved to be a gargantuan effort.
+In order to keep in line with the newly specified https://kb.vshn.ch/app-catalog/reference/framework-requirements.html#_for_service_maintainer["more buy than make"] approach, we're evaluating the viability of using existing operators for the use of converged services.
+
+* Why build on our own when there are already plenty of viable alternatives around?
+* Unpredictable amount of maintenance needed to keep our own solution up-to-date and maintained
+* What do we gain when building our own vs. using an existing solution and integrating it into our systems
 
 == Evaluated Operators
 
 [cols="1,1,1,1,1,1,1"]
 |===
-|Requirements |Percona |Stackgres |CloudnativePG |Zalando |Crunchydata |Bitnami Helmchart
+|Requirements |https://docs.percona.com/percona-operator-for-postgresql/index.html[Percona] |https://stackgres.io/[Stackgres] |https://cloudnative-pg.io/[CloudnativePG] |https://github.com/zalando/postgres-operator[Zalando] |https://access.crunchydata.com/documentation/postgres-operator/v5/[Crunchydata] |https://github.com/bitnami/charts/tree/master/bitnami/postgresql-ha[Bitnami Helmchart]
 |Superuser Access |✅ |✅ |✅ |✅ |✅ |✅
 
-|Database Management |❌ |❌ |❌ |✅ |✅ |❌
+|Database Management footnote:crossplane[Can be added via crossplane-provider-sql] |❌ |❌ |❌ |✅ |✅ |❌
 
-|User Management |❌ |❌ |❌ |✅ |✅ |✅
+|User Management footnote:crossplane[]|❌ |❌ |❌ |✅ |✅ |✅
 
 |Service Metrics |❌ |✅ |✅ |❌ |✅ |✅
 
@@ -23,11 +28,11 @@ Creating our own operator to deploy and manage PostgreSQL proved to be a gargant
 
 |Self Service Backup Restore |✅ |✅ |✅ |✅ |✅ |❌
 
-|Encryption at rest |✅ |✅ |✅ |✅ |✅ |✅
+|Encryption at rest footnote:encryption[This is a platform feature, APPUiO Cloud provides this.]|✅ |✅ |✅ |✅ |✅ |✅
 
-|Connection limit of 25/Gb Memory |✅ |✅ |✅ |❌ |✅ |✅
+|Connection limit of 25/Gb Memory footnote:connections[May not be relevant as all solutions have connection pooling. Should probably be scraped.]|✅ |✅ |✅ |❌ |✅ |✅
 
-|Extension Management |❌ |✅ |❌ |✅ |❌ |✅
+|Extension Management footnote:crossplane[]|❌ |✅ |❌ |✅ |❌ |✅
 
 |Custom PostgreSQL Settings |✅ |✅ |✅ |❌ |✅ |✅
 
@@ -38,16 +43,17 @@ Creating our own operator to deploy and manage PostgreSQL proved to be a gargant
 |Upstream Support |Some support available not clear if for operators |✅ |✅ (via EDB) |❌ |✅ |❌
 |===
 
-https://wiki.vshn.net/display/VHT/PostgreSQL+Operator+Evaluations[More detailed] comparison is in our internal wiki.
+See <<Operators Compared>> for a more detailed comparison.
 
-For the POC on which this decision is based on, we opted to test out Stackgres. It's feature set and general feel were good during the evaluation phase.
+For the POC on which this decision is based on, we opted to test out Stackgres.
+It's feature set and general feel were good during the evaluation phase.
 
 == Pros
 
 * Brings cluster functionality
-* Backups out of the box, incl. PIT
-* Auto deploys Prometheus service monitors
-* Brings Dbops CRs that handle some operations in a k8s native way
+* Backups out of the box, incl. Point-In-Time
+* Automatic deployments of Prometheus service monitors
+* Brings DbOps CRs that handle some operations in a k8s native way
 * Security update schedules can be implemented
 * The operator is maintained by a third party
 * Commercial support available
@@ -56,8 +62,8 @@ For the POC on which this decision is based on, we opted to test out Stackgres. 
 
 * We can't support PostgreSQL version or extensions that the operator doesn't provide
 * Crossplane's limitations, as discussed https://kb.vshn.ch/app-catalog/explanations/decisions/converged-service-impl.html[here]
-* We have to deal with some opinionated decision by the operator developers and maybe need to work around them
-* The operator is maintained by a third party
+* We have to deal with some opinionated decision by the operator developers and maybe need to work around them, or accept them as given
+* The operator is maintained by a third party, but is open source and we could contribute if needed
 
 == Risks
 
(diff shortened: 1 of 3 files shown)


docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
```diff
@@ -0,0 +1,84 @@
+= PostgreSQL Operators
+
+== Problem
+
+Creating our own operator to deploy and manage PostgreSQL proved to be a gargantuan effort. In order to keep in line with the newly specified https://kb.vshn.ch/app-catalog/reference/framework-requirements.html#_for_service_maintainer["more buy than make"] approach, we're evaluating the viability of using existing operators for the use of converged services.
```
tobru, line 5: Maybe state a few more issues we've seen, besides the effort:
* Why build on our own when there are already plenty of viable alternatives around?
* Unpredictable amount of maintenance needed to keep our own solution up-to-date and maintained
* What do we gain when building our own vs. using an existing solution and integrating it into our systems

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+== Evaluated Operators
+
+[cols="1,1,1,1,1,1,1"]
+|===
+|Requirements |Percona |Stackgres |CloudnativePG |Zalando |Crunchydata |Bitnami Helmchart
```
tobru, line 11: Link headings to the corresponding websites

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
```diff
@@ -0,0 +1,84 @@
...
+|Custom PostgreSQL Settings |✅ |✅ |✅ |❌ |✅ |✅
+
+|In-Place Major Upgrade |✅ |✅ |❌ |❌ |❌ |❌
+
+|License |Apache License 2.0 |AGPL-3.0 |Apache License 2.0 |MIT |Apache License 2.0 | Apache License 2.0
+
+|Upstream Support |Some support available not clear if for operators |✅ |✅ (via EDB) |❌ |✅ |❌
+|===
+
+https://wiki.vshn.net/display/VHT/PostgreSQL+Operator+Evaluations[More detailed] comparison is in our internal wiki.
```
tobru, line 41: Please do not link to our wiki, that's not useful for external readers. I'd extract the most important parts of the wiki page and when this page is released, delete the wiki page. Otherwise, I'm sure this would cause friction.

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+For the POC on which this decision is based on, we opted to test out Stackgres. It's feature set and general feel were good during the evaluation phase.
```
tobru, line 43:
```suggestion
For the POC on which this decision is based on, we opted to test out Stackgres.
It's feature set and general feel were good during the evaluation phase.
```

Each sentence belongs to its own line.

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+== Pros
+
+* Brings cluster functionality
+* Backups out of the box, incl. PIT
```
tobru, line 48: PIT?

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
```diff
@@ -0,0 +1,84 @@
...
+* Auto deploys Prometheus service monitors
+* Brings Dbops CRs that handle some operations in a k8s native way
```
tobru, line 50:
```suggestion
* Brings DbOps CRs that handle some operations in a k8s native way
```

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
```diff
@@ -0,0 +1,84 @@
...
+* Security update schedules can be implemented
+* The operator is maintained by a third party
+* Commercial support available
+
+== Cons
+
+* We can't support PostgreSQL version or extensions that the operator doesn't provide
+* Crossplane's limitations, as discussed https://kb.vshn.ch/app-catalog/explanations/decisions/converged-service-impl.html[here]
+* We have to deal with some opinionated decision by the operator developers and maybe need to work around them
```
tobru, line 59: Or accept them as a given

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
```diff
@@ -0,0 +1,84 @@
...
+* The operator is maintained by a third party
```
tobru, line 60: But it's Open Source and we could contribute if needed

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
Kidswiss: Point-In-Time

tobru, line 48:
```suggestion
* Backups out of the box, incl. Point-In-Time
```

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
ccremer: I also favor to integrate the wiki page here.

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+https://wiki.vshn.net/display/VHT/PostgreSQL+Operator+Evaluations[More detailed] comparison is in our internal wiki.
+
+For the POC on which this decision is based on, we opted to test out Stackgres. It's feature set and general feel were good during the evaluation phase.
+
+== Pros
+
+* Brings cluster functionality
+* Backups out of the box, incl. PIT
+* Auto deploys Prometheus service monitors
```
ccremer, line 49:
```suggestion
* Automatic deployments of Prometheus service monitors
```

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
```diff
@@ -0,0 +1,84 @@
...
+Risk:: The Operator becomes unmaintained
+
+Option 1:: We or the community carries the project further
+Option 2:: Migration to another operator
+
+=== CRD Versioning
+
+Risk:: CRDs could contain breaking changes
+
+Option 1:: The maintainer provides conversion webhooks
```
ccremer, line 75: I wouldn't consider this as an option, because conversion webhooks do the magic for us and there'd be nothing to do. Or did I miss something?

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
```diff
@@ -0,0 +1,84 @@
...
+Option 2:: We can handle the changes in the compositions, transparent for the user
+Option 3:: We inform the customers they have to do an action after the upgrade
+
+=== License Changes
+
+Risk:: The licensing of the operator could change
```
ccremer, line 81: Changing the licensing is not the problem itself, resp. it matters which kind of license it's being changed to.
```suggestion
Risk:: The licensing of the operator could change to a hostile license
```

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+Option 1:: Migrate to another operator
+Option 2:: Keep using a version before the change applied
```
ccremer, line 84: Option 2 effectively means that the service is not really supported anymore and it "stalls". Eventually something breaks or newer versions of Postgres don't run anymore.
I would not consider this an option for long-term, rather combine it with option 1, at which point I would just list option 1.

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
```diff
@@ -0,0 +1,84 @@
...
+== Pros
+
+* Brings cluster functionality
+* Backups out of the box, incl. PIT
+* Auto deploys Prometheus service monitors
+* Brings Dbops CRs that handle some operations in a k8s native way
+* Security update schedules can be implemented
+* The operator is maintained by a third party
+* Commercial support available
+
```
ccremer, line 54:
```suggestion
* Upgrades to the operator affect all users at the same time, independent of their maintenance window settings.

```

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
Kidswiss: I've included this, because it mitigates the risk. It's by far the best mitigation for this risk, but maybe not always available.

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
Kidswiss: At least for stackgres this statement is not true. It won't restart anything until explicitly done via a DbOps CR.

ccremer: I wasn't referring to DB restarts, but in general if there are breaking changes (either in API or behavior). I guess this isn't only applicable to Stackgres, but for all operators.

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
ccremer: Sure it mitigates it, but this option is not in our hands. I understand these risk assessment as "what are _our_ action options if it occurs".

Maybe we need to rephrase the risk to `CRDs could contain breaking changes that require a manual upgrade path`

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
Kidswiss: Ah then I suggest to rephrase it:

`Upgrading the operator could inntroduce breaking changes`

To some degree we can handle those changes in the compositions. But a good point we should be aware about.

docs/modules/ROOT/pages/explanations/decisions/postgresql.adoc
```diff
@@ -0,0 +1,495 @@
+= PostgreSQL Operators
+
+== Problem
+
+Creating our own operator to deploy and manage PostgreSQL proved to be a gargantuan effort.
+In order to keep in line with the newly specified https://kb.vshn.ch/app-catalog/reference/framework-requirements.html#_for_service_maintainer["more buy than make"] approach, we're evaluating the viability of using existing operators for the use of converged services.
```
tobru, line 6:
```suggestion
In order to keep in line with the newly specified xref:reference/framework-requirements.adoc#_for_service_maintainer["more buy than make"] approach, we're evaluating the viability of using existing operators for the use of converged services.
```

tobru:
Is it worth keeping the detailed evaluation steps? It clutters the page a lot. If it should be kept, I'd move it to a separate page.

Kidswiss:
> Is it worth keeping the detailed evaluation steps? It clutters the page a lot. If it should be kept, I'd move it to a separate page.

Maybe for the sake of future references?

Kidswiss:
We've reached a decision about using operators for the VSHN Managed PostgreSQL Service.

PR #35, opened 2022-08-23
Add decision for API design

Kidswiss:
## Summary
Decision page for the general API design.

## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
@@ -64,7 +64,7 @@ Advantages::
 Disadvantages::
 
 * We don't really have a "common look"
-* The customer needs to know what the underlying service provider is, even for simple instances
+* The user needs to know some specifics of the underlying service provider.
 * There may only be 1–3 common fields for a given service
 
 === Proposal 2: Separate XRDs with Grouped Fields
--- a/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
@@ -2,7 +2,7 @@
 
 == Problem
 
-This decision is about how we can achieve a "common look" as defined https://kb.vshn.ch/app-catalog/explanations/why_exists.html#_standard_way_to_provide_services[here].
+This decision is about how we can achieve a common look as defined https://kb.vshn.ch/app-catalog/explanations/why_exists.html#_standard_way_to_provide_services[here].
 
 The ideal way would be, that we can re-use the exact same XRD for each set of services.
 For example all implementations for PostgreSQL have exactly the same API for the user.
--- a/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
@@ -256,6 +256,71 @@ Disadvantages::
 * Awkward to use, there needs to be a lot of documentation
 * The complexity of the compositions increases drastically with the parsing rules
 
+=== Proposal 5: Group Specific Settings under One Field
+
+This is a sub-variant of proposal 2.
+Everything that has provider specific naming is grouped under a field called `+providerSpecific+`.
+If a provider doesn't support a given top-level parameter (for example `+maintenance+`), then it's omitted.
+
+.Example Claim
+[source, yaml]
+----
+# Exoscale
+apiVersion: exoscale.appcat.vshn.io/v1
+kind: PostgreSQL
+metadata:
+  name: my-instance
+spec:
+  parameters:
+    name: my-instance
+    maintenance:
+      day: 2
+      hour: 22
+      minute: 0
+    backup:
+      hour: 20
+      minute: 0
+    providerSpecific:
+      version: "14"
+      service:
+        type: pg
+        settings: {}
+        plan: hobbyist-2
+
+# VSHN Converged
+apiVersion: exoscale.appcat.vshn.io/v1
+kind: PostgreSQL
+metadata:
+  name: my-instance
+spec:
+  parameters:
+    name: my-instance
+    maintenance:
+      day: 2
+      hour: 22
+      minute: 0
+    backup:
+      hour: 20
+      minute: 0
+    providerSpecific:
+      version: "14.4"
+      service:
+        settings: {}
+      resources:
+        cpu: 2
+        memory: 2048
+        disk: 200G
+----
+
+Advantages::
+
+* All provider specific fields are grouped
+
+Disadvantages::
+
+* There may only be 1–3 common fields for a given service, shifting everything under the `+providerSpecific+` field
+* Increases the nesting
+
 == Decision
 
 == Rationale
--- a/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
@@ -2,7 +2,7 @@
 
 == Problem
 
-This decision is about how we can achieve a common look as defined https://kb.vshn.ch/app-catalog/explanations/why_exists.html#_standard_way_to_provide_services[here].
+This decision is about how to define "a common look" as mentioned https://kb.vshn.ch/app-catalog/explanations/why_exists.html#_standard_way_to_provide_services[here].
 
 The ideal way would be, that we can re-use the exact same XRD for each set of services.
 For example all implementations for PostgreSQL have exactly the same API for the user.
--- a/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
@@ -1,5 +1,8 @@
 = Decision on API Design
 
+[NOTE]
+This decision page is currently still a WIP.
+
 == Problem
 
 This decision is about how to define "a common look" as mentioned https://kb.vshn.ch/app-catalog/explanations/why_exists.html#_standard_way_to_provide_services[here].
--- a/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
@@ -139,7 +139,7 @@ Advantages::
 Disadvantages::
 
 * The user could get confused by objects with the same kind
-* The customer needs to know what the underlying service provider is, even for simple instances
+* The user needs to know some specifics of the underlying service provider.
 
 === Proposal 3: Composed XRDs
 
--- a/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
@@ -77,7 +77,7 @@ metadata:
 spec:
   parameters:
     name: my-instance
-    postgreSQL:
+    service:
       version: '14'
       type: pg
       settings: {}
--- a/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
@@ -1,0 +1,243 @@
+= Decision on API Design
+
+== Problem
+
+This decision is about how we can achieve a "common look" as defined https://kb.vshn.ch/app-catalog/explanations/why_exists.html#_standard_way_to_provide_services[here].
+
+The ideal way would be, that we can re-use the exact same XRD for each set of services.
+For example all implementations for PostgreSQL have exactly the same API for the user.
+Due to the nature of the product, not all implementations provide the same functionality.
+The set of available features is most likely not the same between most providers.
+Also with Crossplane we're a bit limited on what's possible with translating the source API spec into the target API spec.
+
+== Requirements
+
+* APIs for a set of similar services should feel similar
+* They should only need minimal configuration for a default case
+* We should be able to leverage unique features of the service providers
+
+== Proposals
+
+=== Proposal 1: Use same XRD with provider specific fields
+
+There's exactly one XRD for a set of services.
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
```diff
@@ -0,0 +1,243 @@
+= Decision on API Design
+
+== Problem
+
+This decision is about how we can achieve a "common look" as defined https://kb.vshn.ch/app-catalog/explanations/why_exists.html#_standard_way_to_provide_services[here].
```
tobru, line 5:
```suggestion
This decision is about how we can achieve a common look as defined https://kb.vshn.ch/app-catalog/explanations/why_exists.html#_standard_way_to_provide_services[here].
```

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
```diff
@@ -0,0 +1,243 @@
...
+
+The ideal way would be, that we can re-use the exact same XRD for each set of services.
+For example all implementations for PostgreSQL have exactly the same API for the user.
+Due to the nature of the product, not all implementations provide the same functionality.
+The set of available features is most likely not the same between most providers.
+Also with Crossplane we're a bit limited on what's possible with translating the source API spec into the target API spec.
```
tobru, line 11: That warrants a more detailed reasoning, it doesn't say too much this way...

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
```diff
@@ -0,0 +1,243 @@
...
+
+== Requirements
+
+* APIs for a set of similar services should feel similar
+* They should only need minimal configuration for a default case
+* We should be able to leverage unique features of the service providers
```
tobru, line 17:
```suggestion
* APIs for a set of similar services feel similar
* Only minimal configuration is needed for a default case
* Unique features of a service provider can be leveraged
```

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
```diff
@@ -0,0 +1,243 @@
...
+
+* We can use the same XRD
+* Adding new providers to the list is easy
+
+Disadvantages::
+
+* We don't really have a "common look"
+* The customer needs to know what the underlying service provider is, even for simple instances
+
+=== Proposal 2: Separate XRDs
```
tobru, line 62: Maybe use the same semantics as Crossplane does? At the top-level (`spec.X`) _always_ have the same parameters, and then have a provider specific part like `spec.forProvider.X`.

tobru (changes requested):
A few little things. I'd says it's up for discussion with the Team?

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
Kidswiss: I'm not quite sure what you mean here. Both examples have the same parameters under `spec.parameters.*`?

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
```diff
@@ -0,0 +1,250 @@
...
+== Requirements
+
+* APIs for a set of similar services feel similar
+* Only minimal configuration is needed for a default case
+* Unique features of a service provider can be leveraged
+
+== Proposals
+
+=== Proposal 1: Use same XRD with provider specific fields
+
```
ccremer, line 29: I think over time this looks more like an anti-pattern of Kubernetes.
Consider adding AWS and GCP, eventually you'll end up with maximum 3 common parameters, the rest is very provider-specific and in `.spec.parameters.{aws,gcp,exoscale,vshn}` keys. Might as well create different CRDs for each provider and accept that those 3 common parameters are duplicated.

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
```diff
@@ -0,0 +1,250 @@
...
+Disadvantages::
+
+* We don't really have a "common look"
+* The customer needs to know what the underlying service provider is, even for simple instances
+
+=== Proposal 2: Separate XRDs
+
+This approach will have different XRDs for each provider.
+However, they will follow the same structure, as far as possible.
+All the toplevel fields will be the same, the differences will be apparent in the sub-fields.
```
ccremer, line 73: I would remove this artificial limitation, it doesn't make sense in provider-specific CRD.

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
```diff
@@ -0,0 +1,250 @@
...
+      minute: 0
+    backup:
+      hour: 20
+      minute: 0
+----
+
+Advantages::
+
+* Same look between all providers, deploying PostgreSQL to any backend feels familiar
+* Specifics can be handled in a sensible way
```
ccremer, line 131: I wouldn't focus on "same look" too much. The same look can only apply if various providers share a common set of parameters. But then you'd have to agree on the same terminology. But depending on provider, some term might deviate from the DSL for the provider, maybe even mean the opposite.
As soon as one provider doesn't implement the "common" parameters then this "common look" is already outdated and sets expectations for the user only to be disappointed.

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
```diff
@@ -0,0 +1,250 @@
...
+
+Disadvantages::
+
+* The user could get confused by objects with the same kind
+
+=== Proposal 3: Composed XRDs
+
```
ccremer, line 138: tbh I would remove this proposal as its just not a real option.
Having a base CRDs just to generalize 3 parameters is, IMO, unacceptable for both the user and maintainers.

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
```diff
@@ -0,0 +1,250 @@
...
+Disadvantages::
+
+* "same look" will only apply to the base XRDs
+* Pretty complicated to use for the end-user
+
+=== Proposal 4: Same XRD and Parse the Input
+
+We use the same XRD for every provider.
+Most fields are simply strings, and they will then be parsed in the composition.
```
ccremer, line 188: I don't understand this proposal. How is it different from Proposal 1?

ccremer (commented):
I see XRDs as a way to minimize the set of configurable parameters for a specific service or resource, not as an abstracted "interface" to work with all providers. There are simply too many differences between each.

For S3 object bucket between exoscale and cloudscale it somewhat worked, but only by coincidence due to the rather simple nature of the service. But we already had to use `region` that cloudscale uses as a synonym for `zone` that exoscale uses. It would become confusing if somehow exoscale starts using region within zones and or zones within regions, then the terminology is already confusing.

Generalizing all DB services is a lot more complex and will be very hard if not impossible.
Furthermore, [this "common look"](https://kb.vshn.ch/app-catalog/explanations/why_exists.html#_standard_way_to_provide_services) is hardly specific what it means. For me, having all these services abstracted into CRDs under the `appcat.vshn.io` API group is the common look already.

That said, I like proposal 2 the most. Every field has a meaning and is matching the terminology (DSL) the specific provider is using as well. Every other proposal needs extensive documentation about which field applies to what and which ones do not, even if they are in the API as a generalized parameter for the "common look". I think that's a worse UX than having specific API for specific backend cloud provider.

Kidswiss:
> Furthermore, [this "common look"](https://kb.vshn.ch/app-catalog/explanations/why_exists.html#_standard_way_to_provide_services) is hardly specific what it means. 

This is an attempt to define what this common look should be.

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
Kidswiss: But this is the point of the second proposal.

If you look at the example you see that all parameters concerning the size of the instance are under the key `size` then under this key we use the provider specific naming, like plan, or CPU/Memory/Disk. Depending on what the provider actually understands.

Then you know that all configuration about the size for an instance is under the field `size`, thus giving them a common look. Because it may not be apparent that a `plan` or `profile` or whatever name the provider thinks of, is about sizing.

ccremer: I understood the examples. But if a provider doesn't offer any customization about maintenance for example, then the whole `maintenance` top level key doesn't apply. Or if a service uses some sort of autoscaling and `size` wouldn't matter too.

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
Kidswiss: See above comment. Maybe it's not quite clear what I mean with "common set of parameters". Think of them more of sets of more specific parameters.

For example most providers allow you to configure the service to some degree (version or settings). My proposal now states that all those params concerning the configuration of the service should go under `service`. Then, as a user, you know what these params are for. Same thing for the sizing or backups, etc. that might be common for a given service.

If a provider doesn't implement a given set of params at all, then we omit it.

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
Kidswiss: In that case we simply omit it. Then it's also clear from the API that this isn't supported (yet?) for the given provider.

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
ccremer: I have no strong feelings about grouping some provider-specific settings into top-level parameter groups. But I would avoid having those groups present in the API if they don't apply. I would also avoid renaming fields that are not matching the provider's naming and terminology.
If you think having some same-named group across different DB services suffices as having a "common look", then that's fine by me.

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
ccremer: fine by me to leave it as is.

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
Kidswiss: The difference is, that in proposal 1 there would be some provider specific fields. Here, we don't have any specific fields at all.

ccremer: Does that mean the user has to provide the correct string, resp. yaml/json?

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
Kidswiss: I've added the remark as a disadvantage.

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
Kidswiss: I'd like to keep it in the page, just for completeness' sake.

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
Kidswiss: Yes, exactly.

I've added the source for this idea to the text as well.

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
```diff
@@ -0,0 +1,261 @@
...
+
+Advantages::
+
+* We can use the same XRD
+* Adding new providers to the list is easy
+
+Disadvantages::
+
+* We don't really have a "common look"
+* The customer needs to know what the underlying service provider is, even for simple instances
```
ccremer, line 67: What's the purpose of mentioning "simple instance"? single-replica? how is that different to a "complex instance"?
```suggestion
* The user needs to know some specifics of the underlying service provider.
```

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
```diff
@@ -0,0 +1,261 @@
...
+
+Advantages::
+
+* Same look between all providers, deploying PostgreSQL to any backend feels familiar
+* Specifics can be handled in a sensible way
+
+Disadvantages::
+
+* The user could get confused by objects with the same kind
+* The customer needs to know what the underlying service provider is, even for simple instances
```
ccremer, line 142: same here

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
tobru: My concern is, that `.spec.size` is has different options, depending on the provider used, which might be confusing. E.g. on one we have `.spec.size.plan` and on another we have `.spec.size.cpu`. So my proposal (maybe a new one) would be:

```
apiVersion: exoscale.appcat.vshn.io/v1
kind: PostgreSQL
metadata:
  name: my-instance
spec:
  parameters:
    name: my-instance
    version: '14'
    maintenance:
      day: 2
      hour: 22
      minute: 0
    backup:
      hour: 20
      minute: 0
    providerSpecific:
      service:
        type: pg
        settings: {}
        plan: hobbyist-2
```

This introduces the `providerSpecific` field, which is different for each provider, while the other are always the same.

Kidswiss: Added it as another proposal.

Kidswiss:
I've added another proposal and changed some wording. I'm merging it soon, so that we can have a first round of discussion within the team.

tobru:
>  I'm merging it soon, so that we can have a first round of discussion within the team.

Maybe add an admonition at the top of the page stating that this is under construction?

tobru (approved):

PR #40, opened 2022-11-24
Add decision about how to handle namespaces

Kidswiss:
## Summary

From today's discussion how to handle the namespacing.

## Checklist

- [X] Try to isolate changes into separate PRs (to build a better changelog).
- [X] Categorize the PR by setting a good title and adding one of the labels:
- [X] Link this PR to related issues if applicable.

### DIFF
--- a/docs/modules/ROOT/pages/explanations/decisions/converged-service-impl.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/converged-service-impl.adoc
@@ -1,0 +1,52 @@
+= Converged Service Provisioning Location
+
+== Problem
+
+To provision services converged on the same cluster, there are two approaches:
+
+* Run the resources in the same namespace where the claim originates from
+* Provision managed namespaces where the customer doesn't have access to
+
+== Proposal 1: Managed Namespaces
+
+With managed namespace we provision a new namespace for each new instance.
+The customer doesn't have general access to k8s resources in those namespaces.
+The interaction between the customer and the service is limited to the connection details and APIs provided by the service itself.
+
+Advantages::
+
+* Neatly separated services
+* The services feel like cloud services, where usually no access to the underlying VMs/containers is granted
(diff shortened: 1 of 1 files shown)


glrf:
My 2 cents: I think the decision for the namespacing is solid, but I'm not that happy with just removing an old decision as part of this PR. 

If it's not relevant I'd remove it in an other PR with a clear reasoning why the initial decision was flawed

Kidswiss:
> If it's not relevant I'd remove it in an other PR with a clear reasoning why the initial decision was flawed

Would just a separate commit with an appropriate message in this PR also be okay?

glrf:
> Would just a separate commit with an appropriate message in this PR also be okay?

I can live with it,  but it would be cleaner in a separate PR.

The removal just seems more controversial than the addition. Not all of the reasoning in the older decision is invalid in my opinion.

Kidswiss:
> The removal just seems more controversial than the addition. Not all of the reasoning in the older decision is invalid in my opinion.

That's true... Should I maybe put the old decision in an archive or something?

tobru:
> That's true... Should I maybe put the old decision in an archive or something?

Yes, please preserve the old page and mark it clearly as obsolete. And I'd also urge doing that in a separate PR. This here feels like something different / new.

tobru (changes requested):
The current title of this page is a bit too generic IMHO, it's specifically about where to deploy the service to. I'd even argue to add a third option: "Dedicated Service Cluster" (like we do on SPKS) - although I'm not 100% sure. At least it wouldn't affect the decision, with that, I'm happy :+1:

Kidswiss:
> At least it wouldn't affect the decision, with that, I'm happy 👍

That's true. Service clusters would be something different.

tobru (approved):

glrf (approved):

PR #41, opened 2022-11-25
Update decision for the API design

Kidswiss:
## Summary
There were changes to the API design that haven't been recorded in this decision page. Now it should contain all relevant information about how the APIs for AppCat should look like.

## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
@@ -382,8 +382,8 @@ To minimize confusion when querying the claims. The kind should also contain the
 .Example Kind
 [source, yaml]
 ----
-apiVersion: provider.appcat.vshn.io/v1
-kind: ProviderPostgreSQL
+apiVersion: acmecorp.appcat.vshn.io/v1
+kind: AcmecorpPostgreSQL
 ----
 
 == Rationale
--- a/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
@@ -399,7 +399,7 @@ To achieve the identity and same look for the various providers and services a f
 
 In Kubernetes there's ambiguity if there are multiple CRDs with the same `+kind+` but different `+apiversion+` fields.
 
-Given a K8s cluster with these CRDs:
+Given a Kubernetes cluster with these CRDs:
 
 * `+foo.a.io+`
 * `+foo.b.io+`
--- a/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
@@ -404,5 +404,5 @@ Given a Kubernetes cluster with these CRDs:
 * `+foo.a.io+`
 * `+foo.b.io+`
 
-If a user invokes `+kubectl get foo+` only CRs of type `+foo.a.io+` are returned footnote:ordering[As far as I was able to determine the K8s returns the first API group it finds by alphabetical order.].
+If a user invokes `+kubectl get foo+` only CRs of type `+foo.a.io+` are returned footnote:ordering[As far as we were able to determine, the Kubernetes API server returns the first API group it finds by alphabetical order.].
 This could lead to "missing" CRs from a user perspective
--- a/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
@@ -377,7 +377,8 @@ Additional points we've agreed on:
 * Exposing only needed fields, add more as needed
 * Sensible defaults, minimize required parameters.
 
-To minimize confusion when querying the claims. The kind should also contain the provider as a pre-fix:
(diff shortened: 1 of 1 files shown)


ZuGao (commented):
So we are keeping the provider prefix in the `kind` field?

Kidswiss:
> So we are keeping the provider prefix in the kind field?

Yes.

In practice we'll just need to change the apiversion fields for all DBaaS claims.

ZuGao (approved):
LGTM

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
```diff
@@ -377,6 +377,15 @@ Additional points we've agreed on:
 * Exposing only needed fields, add more as needed
 * Sensible defaults, minimize required parameters.
 
+To minimize confusion when querying the claims. The kind should also contain the provider as a pre-fix:
```
tobru, line 380: IMHO that's much more explanation. When I read that in a few month I don't know anymore what's meant with "To minimize confusion".

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
```diff
@@ -377,6 +377,15 @@ Additional points we've agreed on:
...
+
+.Example Kind
+[source, yaml]
+----
+apiVersion: provider.appcat.vshn.io/v1
+kind: ProviderPostgreSQL
```
tobru, line 386:
```suggestion
apiVersion: acmecorp.appcat.vshn.io/v1
kind: AcmecorpPostgreSQL
```

I feel "provider" is too generic to understand it

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
Kidswiss: I've added more in-depth explanation in the `rationale` section of the decision.

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
```diff
@@ -377,10 +377,32 @@ Additional points we've agreed on:
...
 
 Proposal 2 provides the most flexibility for modelling the various limitations and differences that various providers might have.
 All other proposals are hard definitions what the API should look like and are not a convention.
 This makes it hard to tailor the various provider limitations and differences to fit the definitions.
 
 To achieve the identity and same look for the various providers and services a flexible convention makes the most sense.
+
+=== Avoiding Confusion with Specific Kinds
+
+In k8s there's ambiguity if there are multiple CRDs with the same `+kind+` but different `+apiversion+` fields.
```
tobru, line 400:
```suggestion
In Kubernetes there's ambiguity if there are multiple CRDs with the same `+kind+` but different `+apiversion+` fields.
```

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
```diff
@@ -377,10 +377,32 @@ Additional points we've agreed on:
...
+
+Given a K8s cluster with these CRDs:
```
tobru, line 402:
```suggestion
Given a Kubernetes cluster with these CRDs:
```

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
```diff
@@ -377,10 +377,32 @@ Additional points we've agreed on:
...
+
+* `+foo.a.io+`
+* `+foo.b.io+`
+
+If a user invokes `+kubectl get foo+` only CRs of type `+foo.a.io+` are returned footnote:ordering[As far as I was able to determine the K8s returns the first API group it finds by alphabetical order.].
```
tobru, line 407:
```suggestion
If a user invokes `+kubectl get foo+` only CRs of type `+foo.a.io+` are returned footnote:ordering[As far as we were able to determine, the Kubernetes API server returns the first API group it finds by alphabetical order.].
```

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
```diff
@@ -377,10 +377,32 @@ Additional points we've agreed on:
...
+This could lead to "missing" CRs from a user perspective
```
tobru, line 408:
```suggestion
This could lead int to a "missing" CRs situation from a users perspective.
```

tobru (approved):
Maybe reference https://stackoverflow.com/questions/58061321/kubectl-api-resources-lists-duplicate-resources-with-different-api-groups somewhere?

docs/modules/ROOT/pages/explanations/decisions/api-design.adoc
Kidswiss: > This could lead int to a "missing"

Did you mean "This could lead into a missing"?

Kidswiss:
> Maybe reference https://stackoverflow.com/questions/58061321/kubectl-api-resources-lists-duplicate-resources-with-different-api-groups somewhere?

Not sure if that's helpful, the question in your link talks more about versioning (v1beta1, v1, etc.) than the groups. While the effect is the same, it could confuse people not that familiar with Kubernetes (actually happened during the internal discussion).

PR #47, opened 2022-12-12
Add decision for Redis/Galera

Kidswiss:
## Checklist


docs/modules/ROOT/pages/explanations/decisions/redis_galera.adoc
```diff
@@ -0,0 +1,64 @@
...
+But most of those are in a very early development stage.
+Or don't support the applications we want to provide.
+
+The following section contains the operators that have been looked at.
+We might re-evaluate them in the future.
+
+=== Redis
+
+* https://ibm.github.io/operator-for-redis-cluster/
+** Very alpha
```
susana-garcia, line 21: a question here: why is this operator very alpha? I think we talked about it in the refinement, but now I don't recall it and I just saw the last tag is `v1.2.3`, so I wanted to ask.

docs/modules/ROOT/pages/explanations/decisions/redis_galera.adoc
```diff
@@ -0,0 +1,64 @@
...
+== Decision
+
+Bitnami Helm Charts.
+
+Both Helm Charts can cover the single node use-case and the clustered use-case.
+
+Advantages::
+
+* Can easily be forked and maintained in case of breaking changes
+* Haven been used in SPKS for years now
```
susana-garcia, line 53: typo: Have

docs/modules/ROOT/pages/explanations/decisions/redis_galera.adoc
```diff
@@ -0,0 +1,64 @@
+= Redis and MariaDB/Galera Provisioning
```
tobru, line 1: I'd add a page per service and not mix them in the same decision page. And is it only about provisioning?

docs/modules/ROOT/pages/explanations/decisions/redis_galera.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+This page combines the decisions for deploying Redis, MariaDB and Galera, as they are already in use for SPKS.
```
tobru, line 3: That's not a reason for me to merge that into one document, every service has its own identity and live.

docs/modules/ROOT/pages/explanations/decisions/redis_galera.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+== Problem
+
+We want to re-use proven methods to deploy Redis and MariaDB/Galera.
```
tobru, line 7: Is that the actual problem to solve? That sounds to me like the reasoning _why_. The problem statement should more be like "We need to provide service X on a Kubernetes cluster with the following features: ...". Or something like that...

docs/modules/ROOT/pages/explanations/decisions/redis_galera.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+== Operators
+
+For Redis as well as MariaDB/Galera there are a few operators available.
+But most of those are in a very early development stage.
+Or don't support the applications we want to provide.
```
tobru, line 13: Isn't that already the conclusion? IMHO this section should be judgement-free.

docs/modules/ROOT/pages/explanations/decisions/redis_galera.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+The following section contains the operators that have been looked at.
+We might re-evaluate them in the future.
```
tobru, line 16: Also a conclusion IMHO

docs/modules/ROOT/pages/explanations/decisions/redis_galera.adoc
```diff
@@ -0,0 +1,64 @@
...
+=== Redis
+
+* https://ibm.github.io/operator-for-redis-cluster/
+** Very alpha
+* https://github.com/spotahome/redis-operator
+** Somewhat promising
+** Maybe in the future, for now
+** Evaluation by a third party: https://blog.palark.com/failure-with-redis-operator-and-redis-data-analysis-tools/
+* https://operatorhub.io/operator/redis-operator
+** No sentinel support
```
tobru, line 27: What does "Somewhat promising" mean? Why?
Why "Maybe in the future, for now"?

docs/modules/ROOT/pages/explanations/decisions/redis_galera.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+Bitnami Helm Chart: https://github.com/bitnami/charts/tree/main/bitnami/redis/
```
tobru, line 29: What's that here? In the "Operators" section?

docs/modules/ROOT/pages/explanations/decisions/redis_galera.adoc
```diff
@@ -0,0 +1,64 @@
+= Redis and MariaDB/Galera Provisioning
+
+This page combines the decisions for deploying Redis, MariaDB and Galera, as they are already in use for SPKS.
+
+== Problem
+
+We want to re-use proven methods to deploy Redis and MariaDB/Galera.
+
+== Operators
```
tobru, line 9: Shouldn't we talk about different possibilities here and Operators being one of it?

docs/modules/ROOT/pages/explanations/decisions/redis_galera.adoc
```diff
@@ -0,0 +1,64 @@
...
+* Can easily be forked and maintained in case of breaking changes
+* Haven been used in SPKS for years now
+
+Disadvantages::
+
+* Breaking changes could make the maintenance harder
+
+=== Rationale
+
+Helm Charts are nowhere nearly as complicated as operators if the need for a fork arises.
```
tobru, line 61: I am uncertain if this statement makes sense, e.g. we decided for an operator for PostgreSQL, why should that now be different?

docs/modules/ROOT/pages/explanations/decisions/redis_galera.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+Also in case of breaking changes, we can move the responsibility back to the customer, by coupling the helm chart to specific chart versions.
```
tobru, line 63: What does that mean?

tobru (changes requested):
This needs much more love and work to be approved. Look at https://kb.vshn.ch/app-catalog/explanations/decisions/postgresql.html for example, which is a more thorough documentation. Keep in mind that this is important work for our "future us", to be apparent why we did what and to re-iterate over decisions.

docs/modules/ROOT/pages/explanations/decisions/redis_galera.adoc
Kidswiss: I think you're looking at the wrong one, this one is at 0.3.13.

susana-garcia: oh, right, sorry. All clear, thanks

docs/modules/ROOT/pages/explanations/decisions/redis_galera.adoc
zugao: What @Kidswiss meant here is that having `our own operator` is costly in term of time and effort and we are better off using an existing (or forked) Helm Char.t

Kidswiss: Yeah, `our own` is a good keyword here.

Kidswiss:
I should have marked this as a draft :)

The page will need a lot more love, but the higher prio for me was that we reach a decision already. Because I need to start planning things.

Kidswiss:
@tobru 

I've cleaned the mess a bit, what do you think?

tobru (approved):
OK to me. Before merging, please add a few words about how these Helm Charts are being integrated in our AppCat "framework", aka that we're making use of provider-helm to make these Helm Charts Kubernetes native instantiable.

Please also link to the exact Helm Chart to be used again in the decision. Maybe we should also mention a possible solution to the documented disadvantage?

These new pages are named "VSHN Managed X" whereas the PostgreSQL page is named "PostgreSQL Operators". For consistency, please also rename the PostgreSQL page and reorder the navigation so that they are besides each other.

Kidswiss:
I've added your suggestions, thanks.

PR #54, opened 2023-02-01
Rename Service Requirement iterations

tobru:
The idea with the fantasy names didn't work out, so let's name the iterations what they actually contain.


Kidswiss (approved):
This makes them actually usable 🎉

lieneluksika (approved):
Yesssss!

PR #55, opened 2023-02-07
Rename Service Requirements into Service Maturity

tobru:


docs/modules/ROOT/pages/reference/service-maturity.adoc
```diff
@@ -1,10 +1,11 @@
...
+= Service Maturity
+:page-aliases: service-requirements.adoc
 
 Implementing and maintaining a service is iterative work.
 The idea is to release early and release often, without having the urge to have all features available right from the beginning.
 
-This page gives an idea on the iterative steps.
-Except the initial version and the feature complete version, all steps in-between can be in any order, as it fits best.
+This page gives an idea on the iterative steps and a way to assess the service maturity.
+Except the provisioning maturity and the stable release, all steps in-between can be in any order, as it fits best.
```
lieneluksika, line 8: I am not sure that I understand what is meant by provisioning maturity.

Kidswiss:
```suggestion
Except the "Provisioning" maturity and the stable release, all steps in-between can be in any order, as it fits best.
```

Maybe putting it in quotes and use a capital P helps with the fact that it's a name?

Kidswiss (changes requested):
See my comment below liene's comment.

Kidswiss (approved):

PR #57, opened 2023-02-20
Metrics and Alerting enhancement

lieneluksika:
Moved SLO report from alerting to metrics, added incident prevention by the user to alerting

## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/reference/service-maturity.adoc
+++ b/docs/modules/ROOT/pages/reference/service-maturity.adoc
@@ -58,16 +58,16 @@ See also:
 
 As a user of the service I can:
 
-* see metrics of a service instance
-* see SLO reports
-* access a dashboard showing pre-crafted diagrams
+* see the raw metrics of a service instance
+* see the SLO reports
+* access a dashboard showing pre-crafted diagrams of the available metrics
 
 == Iteration: Alerting
 
 As a user of the service I can:
 
 * enable alerting on pre-defined SLIs to VSHN so that VSHN can resolve incidents with priority
-* enable alerting on pre-defined incident prevention to me so that I can prevent potential incidents
+* enable alerting on the capacity of the instances to an alert channel of my choice so that I can prevent incidents
 
 As an engineer of the service I can:
 
--- a/docs/modules/ROOT/pages/reference/service-maturity.adoc
+++ b/docs/modules/ROOT/pages/reference/service-maturity.adoc
@@ -59,6 +59,7 @@ See also:
 As a user of the service I can:
 
 * see metrics of a service instance
+* see SLO reports
 * access a dashboard showing pre-crafted diagrams
 
 == Iteration: Alerting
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/reference/service-maturity.adoc
```diff
@@ -59,18 +59,19 @@ See also:
 As a user of the service I can:
 
 * see metrics of a service instance
+* see SLO reports
 * access a dashboard showing pre-crafted diagrams
```
lieneluksika, line 63: @Kidswiss could you please explain what exactly is meant by this? Do the 2 upper points possibly cover this?

docs/modules/ROOT/pages/reference/service-maturity.adoc
```diff
@@ -59,18 +59,19 @@ See also:
...
 
 == Iteration: Alerting
 
 As a user of the service I can:
 
 * enable alerting on pre-defined SLIs to VSHN so that VSHN can resolve incidents with priority
-* see SLO reports so that I can get an overview of how well my instance performs
+* enable alerting on pre-defined incident prevention to me so that I can prevent potential incidents
```
tobru, line 70: I have a hard time to understand this sentence. Would you be able to do an iteration over it?

lieneluksika: Right, this is unclear.
The idea is that the customer gets alerts like "storage is getting full", not VSHN. 

Would this be clearer? Using the same wording as we have in the product description. 
"enable alerting on the capacity of the instances to an alert channel of my choice so that I can prevent incidents"

docs/modules/ROOT/pages/reference/service-maturity.adoc
lieneluksika: with this I mean "access a dashboard showing pre-crafted diagrams"

Kidswiss: That the customer has a link to a page where they can see visualizations of the available metrics.

docs/modules/ROOT/pages/reference/service-maturity.adoc
Kidswiss: That sounds much clearer to me, yes.

docs/modules/ROOT/pages/reference/service-maturity.adoc
lieneluksika: Would that not be the same as "see metrics of a service instance"?

Kidswiss: Not necessarily. I interpreted that one that we expose the raw metrics for the customer and they can scrape and visualize themselves. The dashboard with our pre-crafted diagrams would be added bonus so they don't have to do it themselves.

docs/modules/ROOT/pages/reference/service-maturity.adoc
tobru: Perfect :+1:

docs/modules/ROOT/pages/reference/service-maturity.adoc
lieneluksika: That does sound good. Should we add "of the available metrics" at the end of "access a dashboard showing pre-crafted diagrams"? That way it would be clear what the graphs show.

Kidswiss: Yeah sounds good.

I would probably also re-write the other one to "see the raw metrics of a service instance". So it's clear to them that the raw metrics are also available if they wish.

PR #58, opened 2023-02-21
adding decision for postgresql monitoring

wejdross:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
@@ -1,109 +1,51 @@
-= PostgreSQL Managed by VSHN Monitoring
+= PostgreSQL by VSHN Availability check
 
 == Problem
 
-We need to provide PostgreSQL Managed by VSHN monitoring checks:
+We need to monitor availability of PostgreSQL by VSHN, based on:
 
-* Availability check
+* SLA: 99.99%
+* SLO: 99.995%
+* SLI: 99.995%
 
-== Requirements for availability check
+== Requirements for SLI
 
 * PostgreSQL is accessible on network level
-* it's possible to login to PostgreSQL
-* it's possible to execute queries 
+* The end user is able to login and execute database queries
 
 == Solutions
 
-PostgreSQL managed by Stackgres:
+1. Integrated postgresql_exporter into Stackgres cluster
+2. Postgresql_exporter managed and run by VSHN
+3. Customly written exporter
 
-* it doesn't reuse generated connection string which our users will use, which could lead to situation where our monitoring is green, but users can't login (issues with Service)
-* it's sidecar, issue with exporter would require pod restarting
-* prometheusAutobind from SGCluster doesn't work, so I have a fear that whole monitoring isn't really well tested
+== Pros and cons of all solutions
 
-PostgreSQL managed by VSHN:
+"Exporter" should be understood as https://github.com/prometheus-community/postgres_exporter[PostgreSQL_Exporter]
 
-* the same functionalities as above
-* solves issues presented above
+[cols="1,1,1,1"]
+|===
+|Requirements/Options |Exporter provided by VSHN |Exporter provided by Stackgres |Custom exporter
 
-== Decision
+|Checks network connectivity |✅  |✅ |✅
 
-Use postgresql_exporter managed by VSHN, created as Kubernetes Object in our composition:
-```
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: exporter-testing
-  labels:
-    name: exporter-testing
----
-apiVersion: vshn.appcat.vshn.io/v1
-kind: VSHNPostgreSQL
-metadata:
-  name: my-postgres-example
-  namespace: exporter-testing
-spec:
-  parameters:
-    service:
-      majorVersion: "15"
-    backup:
-      schedule: '0 22 * * *'
-    size: 
-      cpu: "300m"
-      memory: "64Mi"
-      disk: "1Gi"
-  writeConnectionSecretToRef:
-    name: postgres-creds-connection
----
-apiVersion: v1
-kind: Pod
-metadata:
-  name: "postgres-exporter"
-  namespace: exporter-testing
-  labels:
-    app: "postgres-exporter"
-spec:
-  containers:
-  - name: postgres-exporter
-    image: "prometheuscommunity/postgres-exporter"
-    command: ["/bin/sh", '-c']
-    args: ["/bin/postgres_exporter --log.level=debug"]
-    resources:
-      limits:
-        cpu: 200m
-        memory: 500Mi
-      requests:
-        cpu: 100m
-        memory: 200Mi
-    env:
-    - name: DATA_SOURCE_URI
-      valueFrom:
-        secretKeyRef:
-          name: postgres-creds-connection
-          key: POSTGRESQL_HOST
-    - name: DATA_SOURCE_USER
-      valueFrom:
-        secretKeyRef:
-          name: postgres-creds-connection
-          key: POSTGRESQL_USER
-    - name: DATA_SOURCE_PASS
-      valueFrom:
-        secretKeyRef:
-          name: postgres-creds-connection
-          key: POSTGRESQL_PASSWORD
-    - name: PG_EXPORTER_AUTO_DISCOVER_DATABASES
-      value:  'true'
-    ports:
-    - containerPort:  9187
-  restartPolicy: Always
----
-# kubectl -n exporter-testing port-forward pods/postgres-exporter 9187
-# curl localhost:9187/metrics | grep -v '#'
-```
-Advantages::
+|Login to database the same way as customer do |✅ |❌ |✅
 
-* the same set of metrics
-* we test connection in the same way as our customers use it
+|PostgreSQL metrics |✅ |✅ |❌
+
+|Easily integrates with prometheus |✅ |❌ |✅
+
+|Read queries check |✅ |✅ |✅
+
+|Write queries check |❌ |❌ |✅
+
+|===
 
 === Rationale
 
-We should go with postgresql_exporter managed by us, not Stackgres, as it solves few issues instead of creating potential ones
+PostgreSQL_exporter comes with many useful metrics we can use later for Grafana charts, also it has `pg_up` metric which tells us that PostgreSQL was able to handle a connection, authenticate user and execute SELECT queries. 
+
+=== Decision
+
+We should use postgresql_exporter which runs in user namespace and reuses user's credential to handle connection and queries execution.
+
--- a/docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
@@ -27,7 +27,7 @@ We need to monitor availability of PostgreSQL by VSHN to be able to measure "Up"
 
 ===  Customly written exporter
 
-    We can write our own exporter which we could run in database namespace. In such scenarion we can write minimalistic exporter that checks connectivity reusing customer connection string, can execute WRITE queries and it'll return for example pg_up metric if connectivity and filesystem are both ok. Such small exporter can be then reused for other services to do the same job for us. 
+    We can write our own exporter which we could run in database namespace. In such scenarion we can write minimalistic exporter that checks connectivity reusing customer connection string, can execute WRITE queries and it'll return for example pg_up metric if connectivity and filesystem are both ok. Such small exporter can be then reused for other services to do the same job for us. The custom exporter allows us to be more flexible in how we handle the exceptions to the availability guarantee, as we can handle them in a general purpose programming language instead of having to model them in PromQL.
 
 [cols="1,1,1,1"]
 |===
--- a/docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
@@ -46,7 +46,7 @@ The custom exporter allows us to be more flexible in what we measure and how we 
 
 |Checks network connectivity |✅ |✅ |✅
 
-|Login to database the same way as customer do |✅ |❌ |✅
+|Login to database the same way as the service user |✅ |❌ |✅
 
 |Easily integrates with Prometheus |✅ |✅ |✅
 
--- a/docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
@@ -58,7 +58,9 @@ The custom exporter allows us to be more flexible in what we measure and how we 
 
 === Rationale
 
-The best solution would be to use customly written exporter, because it'll do exactly what we want with minimal footprint on resources. It disadvantage is:
+The best solution would be to write a custom exporter, because it'll do exactly what we want with minimal footprint on resources.
+
+It disadvantage is:
 
 * cost - we need time to write and maintain code
 
--- a/docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
@@ -19,7 +19,12 @@ We need to monitor availability of PostgreSQL by VSHN to be able to measure "Up"
 
 === PostgreSQL Server Exporter by Stackgres
     
-    Stackgres provides us built-in postgresql_exporter running as a side car in database pod, which contains many useful metrics, but one especially usefull from availability monitoring perspective - pg_up. Pg_up is a metric created when all preconfigured checks SQL queries are properly executed and no one returned an error. Such metric could generate false positives in situation when one of the queries will be dropped by PostgreSQL. This exporter also connects locally, which makes such solution unreliable for us, because our customers connects in different way, connection has to pass through pg bouncer, it has to reuse user credentials and shouldn't generate false positives if one of two pods crash. It doesn't check for WRITE operations, only read operations.
+Stackgres provides us built-in https://github.com/prometheus-community/postgres_exporter[PostgreSQL Server Exporter] that runs as a side car, which exposes multiple useful metrics.
+One especially useful from availability monitoring perspective is `pg_up`. The `pg_up` is a metric is set to `1` if all preconfigured SQL query checks are properly executed and no one returned an error. We could use this metric more or less directly as our "Up" SLI.
+
+One limitation if this metric is that it could generate false positives in situation when one of the queries will be dropped by PostgreSQL. 
+The exporter also connects locally, which makes such solution unreliable for us, because our customers connects in different way, connection has to pass through pg bouncer, it has to reuse user credentials and shouldn't generate false positives if one of two pods crash. 
+It doesn't check for WRITE operations, only read operations.
 
 === Postgresql_exporter by VSHN
 
--- a/docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
@@ -48,7 +48,7 @@ The custom exporter allows us to be more flexible in what we measure and how we 
 
 |Login to database the same way as customer do |✅ |❌ |✅
 
-|Easily integrates with prometheus |✅ |✅ |✅
+|Easily integrates with Prometheus |✅ |✅ |✅
 
 |Read queries check |✅ |✅ |✅
 
--- a/docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
@@ -17,7 +17,7 @@ We need to monitor availability of PostgreSQL by VSHN to be able to measure "Up"
 
 == Solutions
 
-=== Postgresql_exporter by Stackgres
+=== PostgreSQL Server Exporter by Stackgres
     
     Stackgres provides us built-in postgresql_exporter running as a side car in database pod, which contains many useful metrics, but one especially usefull from availability monitoring perspective - pg_up. Pg_up is a metric created when all preconfigured checks SQL queries are properly executed and no one returned an error. Such metric could generate false positives in situation when one of the queries will be dropped by PostgreSQL. This exporter also connects locally, which makes such solution unreliable for us, because our customers connects in different way, connection has to pass through pg bouncer, it has to reuse user credentials and shouldn't generate false positives if one of two pods crash. It doesn't check for WRITE operations, only read operations.
 
--- a/docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
@@ -26,9 +26,12 @@ One limitation if this metric is that it could generate false positives in situa
 The exporter also connects locally, which makes such solution unreliable for us, because our customers connects in different way, connection has to pass through pg bouncer, it has to reuse user credentials and shouldn't generate false positives if one of two pods crash. 
 It doesn't check for WRITE operations, only read operations.
 
-=== Postgresql_exporter by VSHN
+=== PostgreSQL Server Exporter by VSHN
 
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,109 @@
...
+* it's possible to login to PostgreSQL
+* it's possible to execute queries 
+
+== Solutions
+
+PostgreSQL managed by Stackgres:
+
+* it doesn't reuse generated connection string which our users will use, which could lead to situation where our monitoring is green, but users can't login (issues with Service)
+* it's sidecar, issue with exporter would require pod restarting
+* prometheusAutobind from SGCluster doesn't work, so I have a fear that whole monitoring isn't really well tested
```
zugao, line 21: Have you tested with a ServiceMonitor? If it does not bind automatically maybe the ServiceMonitor resource is missing for some reason.

wejdross: I had installed prometheus-operator on the cluster, I set this option to True and there was no ServiceMonitor, no entry in logs that it was/wasn't create, no errors, nothing

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,109 @@
+= PostgreSQL Managed by VSHN Monitoring
```
tobru, line 1:
```suggestion
= PostgreSQL by VSHN Monitoring
```

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+== Problem
+
+We need to provide PostgreSQL Managed by VSHN monitoring checks:
```
tobru, line 5: Should we mention and link to where this requirement comes from? Aka the SLI/SLO/SLA.

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+* Availability check
+
+== Requirements for availability check
+
+* PostgreSQL is accessible on network level
+* it's possible to login to PostgreSQL
```
tobru, line 12:
```suggestion
* The end-user is able to authenticate and get access granted
```

Be as specific as possible here.

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,109 @@
...
+* it's possible to execute queries
```
tobru, line 13:
```suggestion
* The end-user can successfully execute database queries
```

Read only queries or also read/write queries?

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+== Requirements for availability check
+
+* PostgreSQL is accessible on network level
+* it's possible to login to PostgreSQL
+* it's possible to execute queries 
+
+== Solutions
+
+PostgreSQL managed by Stackgres:
```
tobru, line 17: Shouldn't that title say:

```suggestion
Integrated monitoring of Stackgres:
```

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+* it doesn't reuse generated connection string which our users will use, which could lead to situation where our monitoring is green, but users can't login (issues with Service)
```
tobru, line 19:
```suggestion
* It doesn't use the same connection string which the service user uses, which could lead into the situation where our monitoring is green, but the user can't login.
```

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
glrf, line 1: This is really only a decision about measuring uptime, not about monitoring in general

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,109 @@
...
+* it's sidecar, issue with exporter would require pod restarting
```
tobru, line 20:
```suggestion
* The metrics exporter is a sidecar container, an issue with it would require the Pod to restart and therefore causing downtime to the database
```

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
tobru: About wording: Be specific in decision documents. "fear" and "I" do not work in such a document. Be specific and well informed, no guessing is allowed here.

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,109 @@
...
+* prometheusAutobind from SGCluster doesn't work, so I have a fear that whole monitoring isn't really well tested
+
+PostgreSQL managed by VSHN:
```
tobru, line 23: Not sure about the title:

```suggestion
Bring our own metrics exporter:
```

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+* the same functionalities as above
```
tobru, line 25: What functionalities? I only read a few points about issues above, but no functionality.

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,109 @@
...
+* solves issues presented above
```
tobru, line 26: Please be very specific, that doesn't help.

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+== Decision
+
+Use postgresql_exporter managed by VSHN, created as Kubernetes Object in our composition:
```
tobru, line 30: Refer to the solution title once it's clear.

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,109 @@
...
+          key: POSTGRESQL_PASSWORD
+    - name: PG_EXPORTER_AUTO_DISCOVER_DATABASES
+      value:  'true'
+    ports:
+    - containerPort:  9187
+  restartPolicy: Always
+---
+# kubectl -n exporter-testing port-forward pods/postgres-exporter 9187
+# curl localhost:9187/metrics | grep -v '#'
+```
```
tobru, line 101: No technical details needed in a decision document this way. Please remove that again.

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,109 @@
...
+Advantages::
+
+* the same set of metrics
+* we test connection in the same way as our customers use it
```
tobru, line 105: That would go into the "Rationale" part.

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
glrf, line 17: I guess you mean `postgresql_exporter` by Stackgres.

This section is a bit barebone for my taste, most importantly what's missing:

* A proper description what this solution would include. The word `postgresql_exporter` does not come up once in this section
* What metric we'd be using as an uptime measure and why. The exporter provides a lot of metrics. What counts as down and what counts as up?

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+=== Rationale
+
+We should go with postgresql_exporter managed by us, not Stackgres, as it solves few issues instead of creating potential ones
```
tobru, line 109: When I read that I have _no idea_ why we want to go with our own exporter. "It solves this and that, that's why we choose it"?

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
glrf, line 26: Again it's not really clear what this solution would include.

Where would this exporter run? In the customer namespace? In the database namespace? In a separate namespace?

tobru (changes requested):
This document needs a lot more work, this way it's not adhering to a good standard we want to have in our technical decision documents. These documents must clearly help us to understand why we have chosen to do something in a certain way. See a few good examples:

* https://kb.vshn.ch/app-catalog/explanations/decisions/composition-deployments.html
* https://kb.vshn.ch/appuio-cloud/explanation/decisions/control-api.html
* https://kb.vshn.ch/appuio-cloud/explanation/decisions/billing-entity-virtual-resource.html

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+* it doesn't reuse generated connection string which our users will use, which could lead to situation where our monitoring is green, but users can't login (issues with Service)
+* it's sidecar, issue with exporter would require pod restarting
+* prometheusAutobind from SGCluster doesn't work, so I have a fear that whole monitoring isn't really well tested
+
+PostgreSQL managed by VSHN:
+
+* the same functionalities as above
+* solves issues presented above
+
```
glrf, line 27: I can see at least two further options that we should mention and at least explain why we don't want them:

* A pre-exiting query exporter that evaluates SQL queries in defined intervals. Something like: https://github.com/albertodonato/query-exporter (There are more that look similar)
  - That would be closer to measuring the actual user experience than whether the exporter works or not
  - Resolution could be a lot better as we could execute a query every 1-2 second instead of per scrape interval
* A custom SQL query prober.
   - Has same advantages as a pre-existing query prober
   - More flexibility. Could be helpful when we start to integrate all the exception into the SLA reporting
   - Downside is obviously more code we need to maintain

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,109 @@
...
+---
+# kubectl -n exporter-testing port-forward pods/postgres-exporter 9187
+# curl localhost:9187/metrics | grep -v '#'
+```
+Advantages::
+
+* the same set of metrics
+* we test connection in the same way as our customers use it
+
+=== Rationale
```
glrf, line 107: What's IMO missing is any mentioning of measuring SLA exceptions. We don't need to have a solution for all this now, but we should try to think a few steps ahead and this should be a factor for deciding on a solution

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,51 @@
+= PostgreSQL by VSHN Availability check
+
+== Problem
+
+We need to monitor availability of PostgreSQL by VSHN, based on:
+
+* SLA: 99.99%
+* SLO: 99.995%
+* SLI: 99.995%
```
glrf, line 9: A percentage for an SLI doesn't make sense. Also I wouldn't increase the SLO over 99.99%. Having an SLO that is higher than the SLA should allow us to react before we miss the SLA. For such a high SLA that's basically impossible anyway, so there is no point in having an even higher SLO

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,51 @@
...
+== Requirements for SLI
+
+* PostgreSQL is accessible on network level
+* The end user is able to login and execute database queries
+
+== Solutions
+
+1. Integrated postgresql_exporter into Stackgres cluster
+2. Postgresql_exporter managed and run by VSHN
+3. Customly written exporter
```
glrf, line 20: This is still very barebone. What would a customer exporter include? What would a separate postresql_exporter mean?

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,51 @@
...
+
+[cols="1,1,1,1"]
+|===
+|Requirements/Options |Exporter provided by VSHN |Exporter provided by Stackgres |Custom exporter
+
+|Checks network connectivity |✅  |✅ |✅
+
+|Login to database the same way as customer do |✅ |❌ |✅
+
+|PostgreSQL metrics |✅ |✅ |❌
```
glrf, line 34: Not really relevant for the SLI is it?  We definitely want these metrics for other things, but they're not relevant for deciding on an SLI

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
glrf: Still relevant

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,51 @@
...
+
+|Read queries check |✅ |✅ |✅
+
+|Write queries check |❌ |❌ |✅
+
+|===
+
+=== Rationale
+
+PostgreSQL_exporter comes with many useful metrics we can use later for Grafana charts, also it has `pg_up` metric which tells us that PostgreSQL was able to handle a connection, authenticate user and execute SELECT queries.
```
glrf, line 46: The useful additional metrics shouldn't be the main reason to choose an approach. The `pg_up` metric info should be in the solution section. Also there is no rational on why the additional self deployed `postgresql_exporter`, you don't even mention the other solutions.

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,51 @@
...
+
+|===
+
+=== Rationale
+
+PostgreSQL_exporter comes with many useful metrics we can use later for Grafana charts, also it has `pg_up` metric which tells us that PostgreSQL was able to handle a connection, authenticate user and execute SELECT queries. 
+
+=== Decision
+
+We should use postgresql_exporter which runs in user namespace and reuses user's credential to handle connection and queries execution.
```
glrf, line 50: The info about deploying it in the user namespace should be in the solution section.  Also I'm not sure if that's a good idea. It means the user can mess with our metric to measure their SLA.

glrf (commented):
IMO looks a lot better now. 

I have one input for the custom exporter approach. It's not necessarily better, but that's what I envisioned when I thought about the problem.

I have quite a lot of Nitpick about layout and phrasing, but before that I'd like to agree on the decision as a team. So I'd like to have some input from other team members.

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+=== Postgresql_exporter by VSHN
+
+    We can also deploy our own instance of postgresql_exporter to database namespace. We'll receive the same amount of metrics, we'll also be able to measure uptime using pg_up metric, but in this solution we can configure exporter to use the same route, user, password and database as our customers use. It doesn't check for WRITE operations, only read operations.
+
+===  Customly written exporter
+
+    We can write our own exporter which we could run in database namespace. In such scenarion we can write minimalistic exporter that checks connectivity reusing customer connection string, can execute WRITE queries and it'll return for example pg_up metric if connectivity and filesystem are both ok. Such small exporter can be then reused for other services to do the same job for us.
```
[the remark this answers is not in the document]
zugao: That actually is not a bad idea though I expect this to get bigger and bigger as soon as we add more VSHN services. I would prefer this way of dealing with alerts, full control and no compromise on requirements..

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+1. Collect any futher metrics
+2. SLO Definition
+3. Detailed concept of the SLI exceptions
+
+== Solutions
+
+=== Postgresql_exporter by Stackgres
+    
+    Stackgres provides us built-in postgresql_exporter running as a side car in database pod, which contains many useful metrics, but one especially usefull from availability monitoring perspective - pg_up. Pg_up is a metric created when all preconfigured checks SQL queries are properly executed and no one returned an error. Such metric could generate false positives in situation when one of the queries will be dropped by PostgreSQL. This exporter also connects locally, which makes such solution unreliable for us, because our customers connects in different way, connection has to pass through pg bouncer, it has to reuse user credentials and shouldn't generate false positives if one of two pods crash. It doesn't check for WRITE operations, only read operations.
```
glrf, line 22:
```suggestion
Stackgres provides us built-in https://github.com/prometheus-community/postgres_exporter[PostgreSQL Server Exporter] that runs as a side car, which exposes multiple useful metrics.
One especially useful from availability monitoring perspective is `pg_up`. The `pg_up` is a metric is set to `1` if all preconfigured SQL query checks are properly executed and no one returned an error. We could use this metric more or less directly as our "Up" SLI.

One limitation if this metric is that it could generate false positives in situation when one of the queries will be dropped by PostgreSQL. 
The exporter also connects locally, which makes such solution unreliable for us, because our customers connects in different way, connection has to pass through pg bouncer, it has to reuse user credentials and shouldn't generate false positives if one of two pods crash. 
It doesn't check for WRITE operations, only read operations.
```

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+=== Postgresql_exporter by VSHN
+
+    We can also deploy our own instance of postgresql_exporter to database namespace. We'll receive the same amount of metrics, we'll also be able to measure uptime using pg_up metric, but in this solution we can configure exporter to use the same route, user, password and database as our customers use. It doesn't check for WRITE operations, only read operations.
```
glrf, line 26:
```suggestion
=== PostgreSQL Server Exporter by VSHN

We can also deploy our own PostgreSQL Server Exporter instance in the database namespace.

We'll scrape the same metrics, we'll also be able to measure uptime using pg_up metric, but in this solution we can configure exporter to use the same route, user, password and database as our customers use. 
It doesn't check for WRITE operations, only read operations.
```

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+===  Customly written exporter
+
+    We can write our own exporter which we could run in database namespace. In such scenarion we can write minimalistic exporter that checks connectivity reusing customer connection string, can execute WRITE queries and it'll return for example pg_up metric if connectivity and filesystem are both ok. Such small exporter can be then reused for other services to do the same job for us. The custom exporter allows us to be more flexible in how we handle the exceptions to the availability guarantee, as we can handle them in a general purpose programming language instead of having to model them in PromQL.
```
glrf, line 30:
```suggestion
=== Custom Prometheus Exporter

We can write our own exporter that checks connectivity using the same connection details as the service user. It can also execute WRITE queries and it'll return for example pg_up metric if connectivity and file system are both OK.

The exporter can be then reused for other services to do the same job for us.
The custom exporter allows us to be more flexible in what we measure and how we handle the exceptions to the availability guarantee, as we can handle them in a general purpose programming language instead of having to model them in PromQL.
```

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+[cols="1,1,1,1"]
+|===
+|Requirements/Options |Exporter provided by VSHN |Exporter provided by Stackgres |Custom exporter
+
+|Checks network connectivity |✅  |✅ |✅
```
glrf, line 36:
```suggestion
|Checks network connectivity |✅ |✅ |✅
```

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+|Login to database the same way as customer do |✅ |❌ |✅
```
glrf, line 38:
```suggestion
|Login to database the same way as the service user |✅ |❌ |✅
```

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+|Easily integrates with prometheus |✅ |✅ |✅
```
glrf, line 40:
```suggestion
|Easily integrates with Prometheus |✅ |✅ |✅
```

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+|Read queries check |✅ |✅ |✅
+
+|Write queries check |❌ |❌ |✅
+
+|===
+
+=== Rationale
+
+The best solution would be to use customly written exporter, because it'll do exactly what we want with minimal footprint on resources. It disadvantage is:
```
glrf, line 50:
```suggestion
The best solution would be to write a custom exporter, because it'll do exactly what we want with minimal footprint on resources.

It disadvantage is:
```

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+* cost - we need time to write and maintain code
```
glrf, line 52:
```suggestion
* Cost - We need time to write and maintain code
```

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+Its advantages are:
+
+* flexibility - we can extend and adapt code and functionalities as we want to
+* reliability - such exporter reuses customer credentials and connection method so, we're sure customer can connect to the database without issues
+* optimisation - minimal footprint and resource usage
+* operational - we can write it's logic to perform checks each second, this way we can better measure how long PostgreSQL is up or down
+* additional features - compared to postgresql_exporter we can actually check if postgresql is operational by performing write operations on database
```
glrf, line 60:
```suggestion
* Flexibility - We can extend and adapt code and functionalities as we want to.
* Reliability - The exporter reuses service user credentials and connection method so we're sure the service user can connect to the database without issues.
* Optimization - Minimal footprint and resource usage.
* Resolution - We can write its logic to perform checks each second, this way we can better measure how long PostgreSQL is up or down.
* Write Checks - Compared to the PostgreSQL Server Exporter we can actually check if PostgreSQL is operational by performing write operations on the database.
```

docs/modules/ROOT/pages/explanations/decisions/postgres-monitoring.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+=== Decision
+
+We're using custom exporter
```
glrf, line 64:
```suggestion
We're using a custom exporter.
```

glrf (commented):
Mostly formatting and phrasing input.

General input:

* Use paragraphs, it's very hard to read otherwise
* Having one line per sentence makes reviewing easier, otherwise I have multiple input for a single "line" which is actually a whole section

glrf (approved):

zugao (approved):

PR #60, opened 2023-03-08
adding Function docs

wejdross:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/crossplane/working_with_functions.adoc
+++ b/docs/modules/ROOT/pages/how-tos/crossplane/working_with_functions.adoc
@@ -1,0 +1,196 @@
+= How to work with Crossplane Functions
+
+Crossplane introduced `Functions` in version v1.11 as an Alpha feature: https://docs.crossplane.io/knowledge-base/guides/composition-functions. 
+
+Function is a container to which is passed FunctionIO yaml object as stdin. FunctionIO object is created and managed by Crossplane to keep state of our resources. It has 3 main entries:
+
+* observed: this one describes what is current state of our resources, it's read-only, all changes will be discarded
+* desired: this one is state we want to have and it's place where we can add aditional resources or manage already existing ones
+* results: this one is a logger, we can create log entries with 3 levels of severity: Normal, Warning and Fatal - Fatal breaks function's code execution
+
+
+== Preparation step 1
+To start working with Functions we must enable alfa features in Crossplane (at least in version v1.11)
+```
+helm install crossplane --namespace crossplane-system crossplane-stable/crossplane \
+    --create-namespace \
+    --set "args={--debug,--enable-composition-functions}" \
+    --set "xfn.enabled=true" \
+    --set "xfn.args={--debug}"
+```
+
+
+== Preparation step 2
+Next step is to declare our Function in our managed resource (Composition).
+It's highly recommended to use other container registry than hub.docker.io especially for development because their pull limit can block Your dev environment
+```
+spec:
+  compositeTypeRef:
+    apiVersion: vshn.appcat.vshn.io/v1
+    kind: XVSHNPostgreSQL
+  functions:
+    - name: fnio
+      type: Container
+      container:
+        image: ghcr.io/wejdross/fnio
+        imagePullPolicy: Always
+```
+
+
+== Example function 1 - Golang
+
+As a result of code below there will be ConfigMap created even though it wasn't declared in original composition. This code simply takes input from stdin and unmarshal it into xfnv1alpha1.FunctionIO{}, further we append current list of desired resources by our new resource and print it to stdout. You can see that we also create log entry which will print whole aboject we received - be carefull with that, as in my case it was almost 5000 lines. 
+
+```
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+
+	xfnv1alpha1 "github.com/crossplane/crossplane/apis/apiextensions/fn/io/v1alpha1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/yaml"
+)
+
+func main() {
+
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/how-tos/crossplane/working_with_functions.adoc
```diff
@@ -0,0 +1,196 @@
...
+
+Function is a container to which is passed FunctionIO yaml object as stdin. FunctionIO object is created and managed by Crossplane to keep state of our resources. It has 3 main entries:
+
+* observed: this one describes what is current state of our resources, it's read-only, all changes will be discarded
+* desired: this one is state we want to have and it's place where we can add aditional resources or manage already existing ones
+* results: this one is a logger, we can create log entries with 3 levels of severity: Normal, Warning and Fatal - Fatal breaks function's code execution
+
+
+== Preparation step 1
+To start working with Functions we must enable alfa features in Crossplane (at least in version v1.11)
```
Kidswiss, line 13:
```suggestion
To start working with Functions we must enable alpha features in Crossplane (at least in version v1.11)
```

docs/modules/ROOT/pages/how-tos/crossplane/working_with_functions.adoc
```diff
@@ -0,0 +1,196 @@
...
+      type: Container
+      container:
+        image: ghcr.io/wejdross/fnio
+        imagePullPolicy: Always
+```
+
+
+== Example function 1 - Golang
+
+As a result of code below there will be ConfigMap created even though it wasn't declared in original composition. This code simply takes input from stdin and unmarshal it into xfnv1alpha1.FunctionIO{}, further we append current list of desired resources by our new resource and print it to stdout. You can see that we also create log entry which will print whole aboject we received - be carefull with that, as in my case it was almost 5000 lines.
```
Kidswiss, line 42:
```suggestion
As a result of code below there will be ConfigMap created even though it wasn't declared in original composition. This code simply takes input from stdin and unmarshal it into xfnv1alpha1.FunctionIO{}, further we append current list of desired resources by our new resource and print it to stdout. You can see that we also create log entry which will print whole boject we received - be carefull with that, as in my case it was almost 5000 lines. 
```

docs/modules/ROOT/pages/how-tos/crossplane/working_with_functions.adoc
```diff
@@ -0,0 +1,196 @@
...
+* All resources in Composition must be named if You plan to use Functions
+```
+    - name: resource1
+      base:
+        apiVersion: kubernetes.crossplane.io/v1alpha1
+        (...)
+```
+
+* it's much easier to write functions in python - less overhead with parsers and types, You just focus on code
+* debugging Functions is hard, it's worth to print our FunctionIO once to log, save it and then operate on it locally
```
Kidswiss, line 194: I guess we could also use such an approach for testing.

Kidswiss (changes requested):
Found a few typos.

One thing: did you test reading other k8s object? For example the secret in the instance namespace of postgresql or redis. I'm asking because that's one of the blockers for properly providing the `POSTGRES_URL` or `REDIS_URL` fields.

Also, the whole python vs golang should probably be a discussion with the whole team.

docs/modules/ROOT/pages/how-tos/crossplane/working_with_functions.adoc
wejdross: I really like that one :D

Kidswiss: oops :D

docs/modules/ROOT/pages/how-tos/crossplane/working_with_functions.adoc
```diff
@@ -0,0 +1,197 @@
+= How to work with Crossplane Functions
+
+Crossplane introduced `Functions` in version v1.11 as an Alpha feature: https://docs.crossplane.io/knowledge-base/guides/composition-functions. 
+
+Function is a container to which is passed FunctionIO yaml object as stdin. FunctionIO object is created and managed by Crossplane to keep state of our resources. It has 3 main entries:
+
+* observed: this one describes what is current state of our resources, it's read-only, all changes will be discarded
```
zugao, line 7: From what I have read, you can also optionally add another resource not related to the composition or XR. It has to be K8s compliant (kind and apiVersion) and goes straight to config field. I would add this info here as well.

docs/modules/ROOT/pages/how-tos/crossplane/working_with_functions.adoc
```diff
@@ -0,0 +1,197 @@
...
+        namespace: default
+      data:
+        fullURL: "https://google.pl"`
+
+	k8sapproved, err := yaml.YAMLToJSON([]byte(object))
+	if err != nil {
+		log.Fatal("from k8sapproved", err)
+	}
+
+	funcIO.Desired.Composite.Resource.Raw = funcIO.Observed.Composite.Resource.Raw
```
zugao, line 95: I think this is unnecessary. From documentation it seems that all resources are already present in the desired state. Instead of having this line of code I would add a comment stating that the desired resource already matches from the observe.

docs/modules/ROOT/pages/how-tos/crossplane/working_with_functions.adoc
```diff
@@ -0,0 +1,197 @@
...
+RUN go mod download
+COPY main.go /build/
+RUN go build -o /build/fnio
+
+FROM ubuntu:latest
+COPY --from=build /build/fnio /fnio
+ENTRYPOINT [ "/fnio" ]
+```
+
+== Example 2 - Python
```
zugao, line 137: Can you add the initial composition here as well like you did for go example? It should be much clearer what is being updated.

zugao (commented):
There are some things to add but overall good examples.

Kidswiss (approved):

PR #79, opened 2023-08-21
Add how to increase quotas

Kidswiss:
## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ else
 	engine_opts ?= --rm --tty --user "$$(id -u)"
 endif
 
-preview_cmd ?= $(engine_cmd) run --rm --publish 35729:35729 --publish 2020:2020 --volume "${PWD}":/preview/antora ghcr.io/vshn/antora-preview:3.1.2.3 --antora=docs --style=vshn
+preview_cmd ?= $(engine_cmd) run --rm --publish 35729:35729 --publish 2020:2020 --volume "${PWD}":/preview/antora ghcr.io/vshn/antora-preview:3.1.4 --antora=docs --style=vshn
 vale_cmd ?= $(engine_cmd) run $(engine_opts) --volume "$${PWD}"/docs/modules/ROOT/pages:/pages --workdir /pages ghcr.io/vshn/vale:2.15.5 --minAlertLevel=error .
 
 UNAME := $(shell uname)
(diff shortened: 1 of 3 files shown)


docs/modules/ROOT/pages/how-tos/appcat/qppuio-quotas.adoc
```diff
@@ -0,0 +1,35 @@
+= Change AppCat service quota on APPUiO Cloud
+
+It's important, that the service instance has been created with resources lower than the quota.
+This is due to the fact, that the quotas are bound to the namespace and in order to change them, we need the namespace itself.
```
zugao, line 4: This is confusing... why the namespace quote is not automatically scaled based on the service instance? If the customer can change the quote of the namespace then this should happen automatically at service creation.

Kidswiss: The customer can't change the quotas.

This is our internal documentation for increasing the quotas when/if a customer asks us to do that.

zugao (approved):
LGTM

docs/modules/ROOT/pages/how-tos/appcat/qppuio-quotas.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+== Namespace quotas
+The amount of namespaces and thus the amount of AppCat instances that can be created for an organization on APPUiO is by default 25.
```
tobru, line 7: Can you link to external for the number? Should APPUiO Cloud change the value, this documentation will be outdated. Same for every number.

docs/modules/ROOT/pages/how-tos/appcat/qppuio-quotas.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+If that limit is reached, the customer can request an increase in the amount of namespaces they can have.
+This is controlled by a ConfigMap with the name `overrides-$orgName` in the namespace `appuio-cloud`.
```
tobru, line 10: Wouldn't it make more sense to point to the APPUiO Cloud documentation, rather than document APPUiO Cloud specifics here?

tobru (changes requested):

tobru (approved):

PR #102, opened 2024-06-10
Decision on error handling

zugao:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/explanations/decisions/comp-function-error-handling.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/comp-function-error-handling.adoc
@@ -1,0 +1,64 @@
+= Composition Function Error Handling
+
+In March 2024 we had a severe outage of VSHN PostgreSQL instances.
+This was due to several unfortunate events including our poor handling of errors in composition functions.
+This document will discuss in detail the problem we face with composition functions when errors are not properly handled.
+
+== Problems
+
+* We don't manage errors in our composition functions (CFs) and we assume Crossplane will take care of everything.
+* Missing managed resources from desired function when an unexpected error happens can have dire consequences, such as removal of the whole service.
+* Returning early from composition function (CF) due to a warning can lead to missing managed resources.
+
+== Requirements
+
+* A comprehensive guide on how we should deal with error management.
+* A solution to missing managed resources from desired function when errors occur.
+
+== Proposals
+
+Proposal 1:: Manage Errors in CFs
+
+We don't throw fatal errors but use mostly warnings in our composition functions.
+Using warnings and return right away can remove vital parts of a service.
+In order to mitigate this issue we could define the return error (warning or fatal) directly in the runtime.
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/explanations/decisions/comp-function-error-handling.adoc
```diff
@@ -0,0 +1,64 @@
...
+* A comprehensive guide on how we should deal with error management.
+* A solution to missing managed resources from desired function when errors occur.
+
+== Proposals
+
+Proposal 1:: Manage Errors in CFs
+
+We don't throw fatal errors but use mostly warnings in our composition functions.
+Using warnings and return right away can remove vital parts of a service.
+In order to mitigate this issue we could define the return error (warning or fatal) directly in the runtime.
```
Kidswiss, line 24: For the sake of the decision: quickly explain the difference between warning and fatal and the implications. Maybe not everyone knows the difference.

zugao: Added Definition section

docs/modules/ROOT/pages/explanations/decisions/comp-function-error-handling.adoc
```diff
@@ -0,0 +1,71 @@
+= Composition Function Error Handling
+
+In March 2024 we had a severe outage of VSHN PostgreSQL instances.
+This was due to several unfortunate events including our poor handling of errors in composition functions.
+This document will discuss in detail the problem we face with composition functions when errors are not properly handled.
+
+== Problems
+
+* We don't manage errors in our composition functions (CFs) and we assume Crossplane will take care of everything.
```
Kidswiss, line 9: I mean... we do manage errors, but we're not doing a very good job :D

docs/modules/ROOT/pages/explanations/decisions/comp-function-error-handling.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+== Requirements
+
+* A comprehensive guide on how we should deal with error management.
+* A solution to missing managed resources from desired function when errors occur.
+
+== Definitions
+
+* Fatal Result error - the error that stops crossplane from reconciliation which means the desired object will not be applied in the cluster.
+Typically safe move which will ensure no changes to the service but we rarely use this type of error.
```
Kidswiss, line 21: Maybe also mention, that usage of the fatal error at the wrong point will lead to deadlocks. E.g. throwing a fatal error, if the observed object is not yet deployed, which will cause the object to never be deployed at all.

Kidswiss (approved):

docs/modules/ROOT/pages/explanations/decisions/comp-function-error-handling.adoc
zugao: you are right, added `properly`

PR #103, opened 2024-06-12
adding Nextcloud decision

wejdross:
## Summary

* Decision how do we want to deploy Nextcloud

## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/explanations/decisions/nextcloud.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/nextcloud.adoc
@@ -59,4 +59,4 @@ We use the Nextcloud Helm Chart.
 
 == Rationale
 
-As I really like concept of writing our own provider, I think that the Nextcloud Helm Chart is a good starting point for our offering. It is actively maintained and has a lot of features already present. It is also the official Helm Chart for Nextcloud. We can still add features on top of it, like backup and restore, extension management, etc. 
+As I really like concept of writing our own provider, I think that the Nextcloud Helm Chart is a good starting point for our offering. It is actively maintained and has a lot of features already present. It is also the official Helm Chart for Nextcloud. Thanks to composition-functions we can still add features on top of it, like backup and restore, extension management, etc. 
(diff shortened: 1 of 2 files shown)


zugao (commented):
Aren't there more helm chart to choose from? Does the official helm chart cover all aspects of the customer request? If not what else has to be implemented if the helm chart is not enough

wejdross:
> Aren't there more helm chart to choose from? Does the official helm chart cover all aspects of the customer request? If not what else has to be implemented if the helm chart is not enough

As I told You, there is only one project that might be relevant: https://open.greenhost.net/stackspin/nextcloud

but I would go personally with official ones as customer is using the same ones - less problems with integration and migration

There are some other helm charts: https://artifacthub.io/packages/search?ts_query_web=nextcloud&sort=relevance&page=1 but those are mostly dead or one that offers even less features. There are no operators, or Crossplane providers.

docs/modules/ROOT/pages/explanations/decisions/nextcloud.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+* We still need to cover a lot of topics manually:
+** Maintenance
+** Backup and restore
+** Extension management
+** Office package deployment and integration
+
+=== Analysis - Custom Provider
+
+Writing our own provider would give us the most flexibility. We could implement all the features we need and integrate them into our existing solutions. We could probably handle more features than the Helm Chart.
```
Kidswiss, line 45: Creating a custom provider to deploy a few Kubernetes objects seems pretty overkill. Provider-helm can do that much better. We should rather look into forking or creating our helm chart first, if that is necessary.

We've already tried the custom deploy provider approach in the beginning of schedar and it was not feasible. We should build on top and combine things that already exist.

Crossplane providers are usually intended to be very broad in functionality and then through compositions they get the actual business logic applied. So if we would have the need for a provider in the future it should not deploy nextcloud, but rather configure it via its API. It should manage things like:
* user
* extensions
* configurations -> e.g. oauth configs, so that we could automagically attach a vshnkeycloak if it exists
* etc.

There's an important distinction between the responsibilities of a Crossplane provider and a composition. Most providers connect to existing services and provide a means of configuring them. Typically, for each API call there's a single managed resource exposed by the provider. They normally don't provision a service from scratch. The composition then brings the actual business logic and combines the managed resource to so that the service gets created.

If at any point we need to talk to nextcloud's API (for whatever reason) I would strongly suggest not to that in the comp functions but rather move that to the provider-nextcloud. For the reasons stated above.

Kidswiss (approved):
Starting with the only feasible helm chart around sounds good.

Although it might be necessary to go a hybrid approach in the future. Please see my comment about what a provider's responsibility should be.

TheBigLee (approved):

PR #122, opened 2024-11-11
Central Control Plane and Exoscale OSB API

tobru:
## Summary

This document describes how we integrate the VSHN Application Catalog into the Exoscale Marketplace and how the new Central Control Plane works.


docs/modules/ROOT/pages/reference/exoscale-osbapi.adoc
```diff
@@ -0,0 +1,368 @@
...
+    participant VCA as VSHN Control API
+
+    EU->>EP: Enable VSHN Service
+    EP->>VCA: OSB API Provisioning
+    VCA-->>VCA: Create Organization<br/>(if not exist)
+    VCA->>VCA: Create OSBServiceInstance
+    VCA-->>EU: Send E-Mail invitation to Organization<br/>(if not exist)
+    VCA->>EU: Send Service Welcome Mail
+    VCA->>EP: OSB API Confirmation
+    Note over VCA,EP: 201 on successful provisioning<br/>200 if the instance already existed
```
Kidswiss, line 83: By instance you mean the org? Or the actual instance?

docs/modules/ROOT/pages/reference/exoscale-osbapi.adoc
```diff
@@ -0,0 +1,368 @@
...
+We send an E-Mail for each service instance which gets deleted that way, telling the customer that the service has to be removed from the VSHN Portal.
+See also <<Deprovisioning>>, which talks about the details.
+
+The matching `OSBServiceInstance` gets deleted.
+When all `OSBServiceInstance` objects are deleted, an email is sent to customer@vshn.ch for the final closure of the organization.
+
+Also, there is a monitoring check which triggers when no `OSBServiceInstance` is available, but service instances are still there.
+This means the customer didn't delete the service instances.
+
+* TODO: Should we automatically delete service instances after a grace period when there is no `OSBServiceInstance` available?
```
Kidswiss, line 212: IMHO we should set the `OSBServiceInstance` as an owner reference on all relevant objects.

docs/modules/ROOT/pages/reference/exoscale-osbapi.adoc
```diff
@@ -0,0 +1,368 @@
...
+    vshn-- Invoices --> exo
+    exo-- Pays -->vshn
+....
+
+Exoscale must keep track on our pricing on their end, because we only send usage data and they do the calculation.
+
+TODO
+
+* Send billing data to Exoscale billing API - Exoscale does invoicing to customer - we send invoice to Exoscale
+* One SO on VSHN side for Exoscale, to send invoice to Exoscale
```
Kidswiss, line 280: SO?

docs/modules/ROOT/pages/reference/exoscale-osbapi.adoc
```diff
@@ -0,0 +1,368 @@
...
+        {
+            "variable": "something_else",
+            "quantity": 1.2
+        }
+    ]
+}
+----
+<1> `:uuid` is the technical ID of the client organization in the Exoscale backend, which will be shared during the onboarding process.
+
+== OSBServiceInstance
```
Kidswiss, line 339: Wouldn't it make more sense to just track the state of the org with this CRD?

The users can't provision multiple instances via the Exoscale GUI anyway. So a `OSBOrgInstance` is probably enough?

Or is the idea that this is not Exoscale specific? Then I'd go with another name and maybe move it to the `Portal` page. As we've established, Exoscale might be the only one who actually uses OSB, so the OSB specifics could be dropped.

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
```diff
@@ -0,0 +1,162 @@
...
+* A list of available zones
+
+[source,yaml]
+----
+apiVersion: appcat.vshn.io/v1
+kind: OriginConfig
+metadata:
+  name: exoscale
+spec:
+  providerConfigRef:
```
Kidswiss, line 63: Maybe an array? We might have multiple cspcps (omg) per csp

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
```diff
@@ -0,0 +1,162 @@
...
+  XRDRefs:
+    - VSHNPostgreSQL
+    - VSHNMariaDB
+----
+
+This configuration can be overwritten per organization namespace, for example to give access to "beta" services or additional zones.
+
+== Service Catalog
+
+A service is exposed by a https://docs.crossplane.io/v1.18/concepts/composite-resource-definitions/[Crossplane Composite Resource Definitions (XRD)^].
```
Kidswiss, line 77: We should think about having compositions as the single point of truth.

An XRD can implement multiple compositions which in turn could implement different specifics. Like another zone or even provider. That's why `kubectl get appcat` currently queries compositions and not the XRDs.

You even mention below that the true differentiation lies in the composition.

This way we can have a single generic XRD to describe any arbitrary service, including the spec for a claim.
Then the composition will just push that embedded claim to the right control plane.

docs/modules/ROOT/pages/reference/exoscale-osbapi.adoc
```diff
@@ -0,0 +1,368 @@
...
+[mermaid]
+....
+flowchart TB
+    exo["Exoscale Marketplace"]
+    eu("Exoscale End-User")
+    vshn["VSHN Exoscale OSB API"]
+    vshnccpapi["Application Marketplace API"]
+    vshnccpui["Application Marketplace Portal"]
+
+    exo-- OpenServiceBroker APII -->vshn-- K8s APII --> vshnccpapi
```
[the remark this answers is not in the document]
tobru: Yeah, that's crap. It's a hack so that it's rendered correctly. Will solve that issue later :see_no_evil:

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
```diff
@@ -0,0 +1,162 @@
...
+=== Listing
+
+The available service catalog is subject to the organization, and specifically it's origin.
+Therefore, the catalog is made available in the organization's context as a namespaced object called `catalog`.
+
+The object `catalog` is provided by a custom API server, which provides a filtered view on the available <<Service Catalog>>.
+Specifically, it's a filtered list of XRDs.
+It looks up the available XRDs in the <<Origin Metadata>> in the `spec.XRDRefs[]` field.
+
+* TODO: Another way: Have a `catalog` XRD and a controller which reconciles these objects into the Org Namespace, this way we wouldn't need an API server
```
Kidswiss, line 108: In favor of this 👍 API Servers with namespaced objects are a bitch.

docs/modules/ROOT/pages/reference/exoscale-osbapi.adoc
```diff
@@ -0,0 +1,368 @@
...
+    eu-- WebUI -->vshnccpui-- K8s APII --> vshnccpapi
+....
+
+* Our xref:reference/arch-control-plane.adoc[] is used for the heavy lifting, while a custom OSB API endpoint is used for Exoscale integration.
+* The "Enable" button in the Exoscale Portal will:
+** Provision an Organization at VSHN (if it not already exists) which is tied to Exoscale (`origin=exoscale`). See <<Exoscale Organization On- and Offboarding>>
+** Prepare the provisioning of a service instance
+* This Organization will be tied to Exoscale services, it can only see services from Exoscale and only provision at Exoscale
+* On the VSHN Portal, the user can order services, configure and manage them
+* Billing data gets sent to VSHN and Exoscale
```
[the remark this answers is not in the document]
tobru: This part is not clear yet, will figure that out later...

docs/modules/ROOT/pages/reference/exoscale-osbapi.adoc
tobru: This part is completely WIP, just ignore it right now

(SO = Sales Order in Odoo)

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
```diff
@@ -0,0 +1,162 @@
...
+    EU->>CCP: Create Claim
+    CCP->>CCP: Validate Claim
+    CCP->>CSPCP: Create Claim
+    CSPCP->>WRK: Create Instance
+    WRK->>CSPCP: Created
+    CSPCP->>CCP: Created
+    CCP->>EU: Created
+....
+
+* TODO: How are credentials synchronized from the CSPCP to the CCP for the end-user to see?
```
Kidswiss, line 134: Either a push or a pull approach I guess. I feel like we'll need bi-directional connection between the CCP and the CSPCPs.

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
```diff
@@ -0,0 +1,162 @@
...
+* TODO: How to give end-user access to backup listing?
```
Kidswiss, line 135: We'd have to funnel that information from the various service clusters to the CCP. I'm also thinking a controller here that would watch the `kubectl get vshn*backup` objects.

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
Kidswiss: From the discussion here more details how I think this could look (conrete names TBD):

```

apiVersion: vamp.vshn.io/v1
kind: CenrtralServiceClaim # CSC to keep the acronyms :D
metadata:
  name: serviceA #probably generated from something
  namespace: orgNamespace
spec:
  claimContent:
    apiVersion: vshn.appcat.vshn.io/v1
    kind: VSHNPostgreSQL
    metadata:
       name: serviceA
       namespace: orgNamespace
     *RESTOFCLAIM* 
  compositionRef:
    name: providerA-zoneB-serviceName.vamp.vshn.io
  writeConnectionSecretToRef:
    name: serviceA
    namespace: orgNamespace
```

The composition `providerA-zoneB-serviceName.vamp.vshn.io` will then:
* Create a kube object with the right `providerConfig` for that CSPCP
* Maybe even ensure that the org namespace exists in that CSPCP 
* Provide the connection details for that instance back on the CCP

Kidswiss: Thinking even further for the future:

If we're at a point where we have even multiple CSPCPs for a signle zone in for a single CSP, we could then make the `providerA-zoneB-serviceName.vamp.vshn.io` smarter and do actual loadbalancing. E.g. it would connect to the CSPCPs via kubeclient and decide according to some criteria which CSPCP to use.

tobru: > From the discussion here more details how I think this could look (conrete names TBD):
> 
> ```
> 
> apiVersion: vamp.vshn.io/v1
> kind: CenrtralServiceClaim # CSC to keep the acronyms :D
> metadata:
>   name: serviceA #probably generated from something
>   namespace: orgNamespace
> spec:
>   claimContent:
>     apiVersion: vshn.appcat.vshn.io/v1
>     kind: VSHNPostgreSQL
>     metadata:
>        name: serviceA
>        namespace: orgNamespace
>      *RESTOFCLAIM* 
>   compositionRef:
>     name: providerA-zoneB-serviceName.vamp.vshn.io
>   writeConnectionSecretToRef:
>     name: serviceA
>     namespace: orgNamespace
> ```
> 
> The composition `providerA-zoneB-serviceName.vamp.vshn.io` will then:
> 
>     * Create a kube object with the right `providerConfig` for that CSPCP
> 
>     * Maybe even ensure that the org namespace exists in that CSPCP
> 
>     * Provide the connection details for that instance back on the CCP

Doesn't that break the API compatibility? As an end-user, I want to put the same `VSHNPostgreSQL` in the CCP like I do in the CSPCP. By choosing the `compositionRef` I tell the CCP where I would like to make this service happen.

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
```diff
@@ -0,0 +1,399 @@
...
+  writeConnectionSecretsToNamespace: syn-crossplane
+----
+<1> We want to transfer the whole manifest - unclear as of now how to access it
+
+As we need several combinations of Compositions where only the CSPCP connection details are different they are generated via a Project Syn Commodore Component.
+
+=== Listing
+
+The available service catalog is subject to the organization, and specifically it's origin configuration.
+Therefore, the catalog is made available in the organization's context as a namespaced object called `catalog`.
```
simu, line 220: Given the rest of this paragraph, you probably want to use the word "resource" rather than "object" here (generally when referring to a K8s object kind, "resource" is used):

```suggestion
Therefore, the catalog is made available in the organization's context through multiple objects of a namespaced resource called `catalog`.
```

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
```diff
@@ -0,0 +1,399 @@
...
+
+A CRD `Catalog` is provided. Instances of this object are reconciled into the organization namespaces.
```
simu, line 222: Side-note: while it makes sense in context (after trying to parse the text for 10 minutes, see comment on the previous paragraph), this CRD feels misnamed to me. Generally, I'd expect a `Catalog` resource to contain all the services that are available for a certain entity (e.g. a catalog of all Postgres services, or a catalog of all services at Exoscale de-fra-1 or even a catalog of all services that my organization has access to). However, here, a `Catalog` is metadata for a single service (e.g. VSHN Postgres at Exoscale de-fra-1).

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
```diff
@@ -0,0 +1,399 @@
...
+
+This allows to query the available services using `kubectl -n myorg get catalog`.
+
+The catalog reconciler reads the <<Organization Origin Configuration, OrganizationOriginConfig>> for catalog configuration, according to the Organization origin found on the label of the Namespace.
+
+[mermaid]
+....
+flowchart TD
+    A[Start] --> B[Get all namespaces with label<br/>appuio.io/resource.type=organization]
```
simu, line 231: Rendering of this looks broken -> the label details aren't visible in `make preview`

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
```diff
@@ -0,0 +1,399 @@
...
+
+image::universal-control-plane.drawio.svg[]
+
+== Summary
+
+The Central Control Plane (CCP) provides the multi-tenant multi-service-provider aggregation and central entry-point.
+All access to the Meta VSHN Application Marketplace happens through this control plane.
+It offers features needed to separate organizations, control access, service provisioning and gives a central view over multi cloud providers and the instances running this way.
+
+We inherit the original architecture of the https://kb.vshn.ch/appuio-cloud/references/architecture/control-api.html[APPUiO Control API^] and extend it for the CCP.
```
simu, line 15: Is the goal that the CCP replaces the APPUiO Control API, or will these two APIs exist in parallel?

if the goal is to have the CPP replace the APPUiO Control API, there's a bunch of open questions: e.g. how to restrict organizations which are created via e.g. Exoscale Marketplace from creating namespaces on the APPUiO zones? How to restrict APPUiO organizations from creating services on the CSP worker clusters?

simu (commented):
I don't see any obvious issues in the Exoscale OSB integration, but I'm not very familiar with that part.

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
tobru: > Is the goal that the CCP replaces the APPUiO Control API, or will these two APIs exist in parallel?

It will replace the APPUiO Control API. Or in other words: It will advance the APPUiO Control API into a "VSHN Multi-CSP Self-Service API". In essence, APPUiO will become a product in the catalog "Namespace as a Service by VSHN aka APPUiO".

You're right, this part is missing in the docs, I need to dig a bit into what this means for APPUiO specifics today.

simu: > In essence, APPUiO will become a product in the catalog "Namespace as a Service by VSHN aka APPUiO".

It's not clear to me that we can simply make APPUiO namespaces ("Namespace as a Service by VSHN") a product that functions the same as other marketplace services, since currently the base design of APPUiO is implemented as "make it as close to vanilla K8s as possible" and allows users to create arbitrary namespaces directly on the APPUiO zones without interacting with the control-api at all.

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
```diff
@@ -0,0 +1,399 @@
+= Central Control Plane (CCP)
+
+This page discusses the architecture of the Central Control Plane (CCP).
+
+* TODO: Where to store consensus to GTC?
```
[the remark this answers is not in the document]
tobru: Absolutely. I removed this line as it's not needed anymore.

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
tobru: Thanks for bringing this up. You're right, Catalog is not a good name for this kind. I'm trying it with ServiceOffering now, let's see if this resonates better.

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
tobru: Solved by rendering as PNG instead of SVG :man_shrugging:

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
```diff
@@ -0,0 +1,399 @@
...
+  defaultBillingEntityRef: "be-42" <3>
+  mandatoryLegalAcceptance: "true" <4>
+----
+<1> Label selector to filter the available services in the organization origin
+<2> Require `ProvidedService` to enable access to services
+<3> An optional reference to a https://kb.vshn.ch/appuio-cloud/references/architecture/control-api-billing-entity.html[`BillingEntity`]
+<4> When set to true, the Organization needs to accept the legal terms before it is allowed to use the platform. TODO via "Organization" object or via "BillingEntity"? Where should it land in the end? In VSHN Central?
+
+When a `spec.defaultBillingEntityRef` is set, it automatically sets this value on the `Organization` object at `spec.billingEntityRef`.
+Changing the `spec.billingEntityRef` field in the `Organization` is denied, except by the reconciliation loop of `OrganizationOriginConfig`.
```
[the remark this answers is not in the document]
tobru: That makes sense. I updated the document accordingly.

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
tobru: I push this part to be detailed out a bit into the near future. For now, I focus on the Marketplace part. How we proceed (technically) with the APPUiO case will be handled in a follow-up research.

Currently, I think this will end-up similar to what you wrote: It will be handled as selective access to APPUiO clusters (zones). When access is granted, it continues to function as today.

docs/modules/ROOT/pages/reference/exoscale-osbapi.adoc
```diff
@@ -0,0 +1,452 @@
...
+* _Service Offering_: The advertisement of a Service that a Service Broker supports.
+* _Service Plan_: The representation of the costs and benefits for a given variant of the Service Offering, potentially as a tier.
+* _Service Instance_: An instantiation of a Service Offering and Service Plan.
+
+Exoscale Marketplace::
+The https://www.exoscale.com/marketplace/[marketplace^] of Exoscale enabling the self-service ordering of services from third-party providers.
+
+== Architecture
+
+[mermaid]
```
Kidswiss, line 35: Maybe switch this to png as well, it takes ages to load.

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
```diff
@@ -0,0 +1,397 @@
...
+                kind: VSHNPostgreSQL
+            providerConfigRef:
+                name: exoscale-de-fra-1
+        patches:
+        - type: FromCompositeFieldPath
+          fromFieldPath: "TODO" <1>
+          toFieldPath: spec.forProvider.manifest
+  writeConnectionSecretsToNamespace: syn-crossplane
+----
+<1> We want to transfer the whole manifest - unclear as of now how to access it
```
Kidswiss, line 213: We can copy `metadata` and `spec` as a whole in two patch operations.

Something like this:

```yaml
patches:
        - type: FromCompositeFieldPath
          fromFieldPath: spec
          toFieldPath: spec.forProvider.manifest.spec
       - type: FromCompositeFieldPath
          fromFieldPath: metadata
          toFieldPath: spec.forProvider.manifest.metadata
```

docs/modules/ROOT/pages/reference/arch-control-plane.adoc
```diff
@@ -0,0 +1,397 @@
...
+
+[source,yaml]
+----
+apiVersion: vamp.vshn.io/v1
+kind: OrganizationOriginConfig
+metadata:
+  name: exoscale
+spec:
+  compositionSelector: <1>
+    metadata.appcat.vshn.io/serviceprovider: exoscale
```
Kidswiss, line 125: Maybe this should be an array of `matchlabels`? Each selector will then be `OR`'d.

For example if an org has access to exoscale and cloudscale it would look something like this:
```
compositionSelector:
    - matchLabels:
         metadata.appcat.vshn.io/serviceprovider: exoscale
    - matchLabels:
         metadata.appcat.vshn.io/serviceprovider: cloudscale
```

This will then show both providers.

Kidswiss (changes requested):
Found a few more things. But all in all it should be good.

We'll most definitely find the rest of the issues during implementation.

Kidswiss (approved):
LGTM

PR #125, opened 2024-12-05
Add docs about SLI Prober with split control plane

Kidswiss:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/reference/sli-prober-architecture.adoc
+++ b/docs/modules/ROOT/pages/reference/sli-prober-architecture.adoc
@@ -1,0 +1,36 @@
+= SLI Prober
+
+The SLI-Prober connects to each instance and will report on its SLI.
+For that it needs to reconcile the composites of each instance, to get the connection details.
+This way we ensure that the prober probes exactly what the customer gets.
+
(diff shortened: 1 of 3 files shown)


docs/modules/ROOT/pages/reference/sli-prober-architecture.adoc
```diff
@@ -0,0 +1,35 @@
...
+This way we ensure that the prober probes exactly what the customer gets.
+
+To avoid issues with potentially checking thousands of instances with a single prober, we deploy the prober to each service cluster.
+
+In order for the prober to work correctly it needs to reconcile two types of resources:
+
+* The composites
+* `UpgradeJobs`
+
+The composites are reconciled to detect all the actual instances and their connection details.
```
zugao, line 14: Can you also add here how the prober figures out which composites are to be reconciled and which are to be discarded? If there are 2 service clusters connected to one control plane then assume half of instances must be discarded from on SLI prober service.

Kidswiss: That's a good point, actually. I think the current logic would probably fall over. Need to look into that.

PR #128, opened 2024-12-17
initial version of odoo hosting

tobru:


janikvonrotz:
Thanks for the spec.

If it is variant 1 or 2 I guess depends on the configuration options.

I have further updated the Odoo image: https://odoo.build/image/
This should answer the Todos about the origin of the image, performance configuration and how addons are handled.

Regarding the deployment options, I'm working on an Helm repo that contains an Odoo chart. For the development environment, I already made some tests: https://github.com/Mint-System/Odoo-Build/tree/16.0/k8s

What is not discussed yet, is if both Community and Enterprise should be provided. Choosing either has an impact on the module/feature availability, pricing and most important the upgrade options.

I tried to give a big picture of licensing, hosting and upgrade here: https://wiki.mint-system.ch/odoo-lizenzierung_-hosting-und-upgrade.html

What I would like to provide in the long run, is an upgrade service similar to Odoo.sh. A customer could select a newer Odoo version for the integration environment and the upgrade scripts are automatically applied.

PR #130, opened 2024-12-19
Web Portal Specification

tobru:
This document specifies the web portal which sits in front of the CCP.


tobru:
This is now at https://docs.servala.com/web-portal.html

PR #133, opened 2025-01-30
Add ADR for Function Revisions

Kidswiss:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
+++ b/docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
@@ -13,6 +13,8 @@ include::partial$adr-meta.adoc[]
 .Summary
 ====
 In order to make rollouts easier, we want to make sure any changes will only be applied to existing instances during their respective maintenance window.
+
+This ADR assumes that you've read: https://docs.crossplane.io/latest/concepts/composition-revisions
 ====
 
 PoC URL: https://github.com/vshn/component-appcat/pull/612
@@ -25,11 +27,11 @@ Currently, if we have changes that trigger restarts, then we have either to:
 
 Both of these solutions are manual and prone to error. Additionally, the first approach breaks our Renovate process that deploys a change on test clusters for a week before its rolled out everywhere.
 
-Crossplane has a concept called https://docs.crossplane.io/latest/concepts/composition-revisions/#using-composition-revisions[Composition Revisions]. They are automatically created each time a change in a composition is applied. By default, any claim or composite deployed will automatically point to the most recent revision as soon as they are available.
+Crossplane has a concept called https://docs.crossplane.io/latest/concepts/composition-revisions/#using-composition-revisions[Composition Revisions]. They are automatically created each time a change in a composition is applied. By default, any claim or composite deployed will automatically point to the most recent revision as soon as they are available. AppCat currently doesn't make any use of composition revisions at all. This ADR will introduce and extend this concept to AppCat and its Compositon Functions.
 
-This behaviour can be switched to `Manual`, which means that a claim or composite won't automatically update to the latest revision. They are effectively pinned to a version. If a claim or composite gets created with the `Manual` policy, it will point to the latest available revision at the time of creation.
+The behaviour can be switched to `Manual`, which means that a claim or composite won't automatically update to the latest revision. They are effectively pinned to a version. If a claim or composite gets created with the `Manual` policy, it will point to the latest available revision at the time of creation.
 
-Unfortunately these revisions don't work natively with https://docs.crossplane.io/latest/guides/write-a-composition-function-in-go/[Compostion Functions]. Even though there are `functionRevisions` similar to the `compositionRevisions` they can't be used for anything at the moment.
+Unfortunately these revisions don't work natively with https://docs.crossplane.io/latest/guides/write-a-composition-function-in-go/[Compostion Functions]. Even though there are `functionRevisions` similar to the `compositionRevisions` they can't be used for anything at the moment. There's an open feature request about this https://github.com/crossplane/crossplane/issues/6139[on Crossplane's github].
 
 == DIY Function Revisions
 It is possible to leverage the current Compositon Revisions together with Composition Functions, however some changes are necessary.
--- a/docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
+++ b/docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
@@ -1,0 +1,176 @@
+= ADR 0030 - Function Revisions
+:adr_author:    Simon Beck
+:adr_owner:     Schedar
+:adr_reviewers:
+:adr_date:      2025-01-29
+:adr_upd_date:  2025-01-29
+:adr_status:    draft
+:adr_tags:      comp-functions,crossplane
+
+include::partial$adr-meta.adoc[]
+
+[NOTE]
+.Summary
+====
+In order to make rollouts easier, we want to make sure any changes will only be applied to existing instances during their respective maintenance window.
+====
+
+PoC URL: https://github.com/vshn/component-appcat/pull/612
+
+== Context
+Currently, if we have changes that trigger restarts, then we have either to:
+
+* time them during the cluster maintenance
+* disable ArgoCD and manually apply Maintenance hooks to re-enable it again
+
+Both of these solutions are manual and prone to error. Additionally, the first approach breaks our Renovate process that deploys a change on test clusters for a week before its rolled out everywhere.
+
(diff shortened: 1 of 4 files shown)


docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+[mermaid,flow,png]
+....
+flowchart TD
+    start[Claim/Composite created] --> test{Test OpenShift Cluster?}
+    test --> |Yes| testcluster[Always use latest revision]
+    testcluster --> stop
+    test --> |No| prod[Set revision policy to manual]
+    prod --> maint[Maintenance Cronjob triggers]
+    maint --> first{Is first maintenance for instance?}
```
zugao, line 97: What kidn of maintenance it is? instance maintenance? the one that is specified in the claim, right?

docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
```diff
@@ -0,0 +1,123 @@
...
+    first --> |Yes| autpolicy[Set revision policy to auto]
+    first --> |No| selector[Set revision selector to most current revision]
```
zugao, line 99: I don't understand this logic when it's NO, it has to go back to policy manual and remain on the older version until its maintenance is triggered and then it can go to the happy path yes and set to auto no?

docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
```diff
@@ -0,0 +1,123 @@
...
+    selector --> changes[Instance reconciles against latest function and receives changes]
+    autpolicy --> selector
+    changes --> stop
+....
+
+For production clusters the default revision policy will be set to manual. This will ensure that newly created instances will just have the most current revision. Test cluster will get the default policy automatic, so they get changes immediately. This will help detect issues.
```
zugao, line 105: > This will ensure that newly created instances will just have the most current revision
Isn't it the opposite? It will make functions(revisions) available, though the instances will not be set to the latest right away, only during the maintenance.

docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+During the very first maintenance for any given instance, the maintenance job will switch the revision policy over to automatic. It will then also set the revision to the must current using a `selector`. The policy has to be set back to automatic in order for the selector to actually apply otherwise it will still keep using the old revision.
```
zugao, line 107:
```suggestion
During the very first maintenance for any given instance, the maintenance job will switch the revision policy over to automatic. It will then also set the revision to the most current using a `selector`. The policy has to be set back to automatic in order for the selector to actually apply otherwise it will still keep using the old revision.
```

zugao: > It will then also set the revision to the most current using a `selector`
Who or better what will set the selector to the necessary revision? Can you please add that here

docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+=== Hotfixes
+It might be necessary that a hotfix has to be applied immediately to all instances.
+
+To handle this an additional job will be specified in the component. This job will have a parameter which contains the explicit version that should be applied to all claims or components.
```
zugao, line 112: Can you add more about how this job is to be added? maybe in the CD GitHub pipeline where during creating a tag the pipeline understands it's a hotfix and generates the necessary job?

zugao (commented):
Some points to discuss

docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
Kidswiss: Yes the instance maintenance.

docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
Kidswiss: No, if it's not the first maintenance, then it's already set to automatic, so we can skip settings the policy. We only have to change from manual to automatic on the first maintenance, when we set the selector for the revision.

Just for the record, here's the rendered chart:
image

docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
Kidswiss: If we create a claim or composite with the manual policy, it will point to the most current revision at the time of creation. If there are new revisions after that point it won't switch to other revisions as long as manual is set.

From the linked Crossplane docs:
> The below XR uses the Manual policy. When this policy is used the XR will select the latest CompositionRevision when it’s first created, but must manually be updated when you wish it to use another CompositionRevision.

docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
Kidswiss: `It` references the maintenance job mentioned in the previous sentence. How should I write it, that it's clear.

docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
Kidswiss: This should not be automatically determined.

I'll add more how it should work.

docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
zugao: I guess I just need some hands on experience :)

docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+== Context
+Currently, if we have changes that trigger restarts, then we have either to:
+
+* time them during the cluster maintenance
+* disable ArgoCD and manually apply Maintenance hooks to re-enable it again
+
+Both of these solutions are manual and prone to error. Additionally, the first approach breaks our Renovate process that deploys a change on test clusters for a week before its rolled out everywhere.
+
+Crossplane has a concept called https://docs.crossplane.io/latest/concepts/composition-revisions/#using-composition-revisions[Composition Revisions]. They are automatically created each time a change in a composition is applied. By default, any claim or composite deployed will automatically point to the most recent revision as soon as they are available.
```
tobru, line 28: Are we using composition revisions already today? I think we should point out this fact in this description. Because if we aren't already using them, I'd point out very prominently that this design introduces the usage of them. Right now, I only ready "between the lines" that we're introducing that concept because of function revisions.

docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+This behaviour can be switched to `Manual`, which means that a claim or composite won't automatically update to the latest revision. They are effectively pinned to a version. If a claim or composite gets created with the `Manual` policy, it will point to the latest available revision at the time of creation.
+
+Unfortunately these revisions don't work natively with https://docs.crossplane.io/latest/guides/write-a-composition-function-in-go/[Compostion Functions]. Even though there are `functionRevisions` similar to the `compositionRevisions` they can't be used for anything at the moment.
```
tobru, line 32: Is there a GitHub issue or a design document which discusses functionRevisions? So that we can track what's going on in the Crossplane project regarding that topic. If there is nothing, I suggest opening a GitHub issue once we merged this ADR and point from the GH issue to this ADR, describing how we're currently solving it.

tobru (changes requested):
I like the concept, seems to make sense to me. Just some small hints, but nothing blocking.

docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
Kidswiss: We're not using them as of now. I'll add something to state that more explicit.

docs/modules/ROOT/pages/adr/0030-function-revisions.adoc
Kidswiss: I actually opened one, but apart from a few other people being interested in that, nothing much came from it yet: https://github.com/crossplane/crossplane/issues/6139

I'll add the link as well.

Kidswiss:
We should also add some more use-cases to this ADR. As from our first tech talk in the team:

We should enhance it with the ability to not only to keep separate functions and revisions for the tags, but also for a list of configurable branches.

So multiple branches and their functions could be rolled out at the same time on the same cluster. This will at least solve some of the bottlenecks we have with the lab.

wejdross (approved):

tobru (approved):

PR #134, opened 2025-02-12
Adapt Forgejo Architecture docs to implementation

tobru:
## Summary

During implementation, we learned a few deficiencies in the architecture, this PR adapts to the actual implementation.


tobru:
@wejdross Does anything else deviate from the implementation which needs to be reflected in the architecture docs?

wejdross (approved):
looks good to me

PR #135, opened 2025-02-12
Servala Web Portal High-Level Architecture

tobru:
This PR contains the high-level architecture for the Servala self-service portal.


tobru:
This is now at:

* https://docs.servala.com/exoscale-osb.html
* https://docs.servala.com/web-portal.html

PR #144, opened 2025-06-02
ADR for billing system

zugao:
## Summary

This ADR presents solutions for our new Event-Based BIlling

## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
+++ b/docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
@@ -1,0 +1,242 @@
+= ADR 0033 - Event-Based Billing
+:adr_author:    Gabriel Saratura
+:adr_owner:     Schedar
+:adr_reviewers:
+:adr_date:      2025-06-02
+:adr_upd_date:  2025-06-02
+:adr_status:    draft
+:adr_tags:      billing,odoo
+
+include::partial$adr-meta.adoc[]
+
+[NOTE]
+.Summary
+====
+For sending billing events to Odoo (https://docs.central.vshn.ch/event-billing.html[Event Based Billing^]), will use the existing AppCat controller to manage billing events and Kubernetes CRDs as the persistence mechanism.
+====
+
+== Requirements
+
+. Reuse existing tools and/or components of AppCat to build a resilient event-based billing solution.
+. Make data resending to Odoo possible in case it's required (historical record retention is required).
+. Persist the state of each billing request.
+. Do not block service creation, deletion, or update due to errors in the billing system.
+. [[odoo-sync-state]] Provide a way to verify sync state between Odoo and the Kubernetes cluster.
+. Do not lose any event (created, deleted, scaled), as this has a direct financial impact.
+
+
+== Solution Options
+
+After careful evaluation, the two most promising solutions for implementing event-based billing are:
+
+. <<AppCat Controller>>
+. <<Runtime Library via Crossplane Composition Functions>>
+
+=== AppCat Controller
+
+Transitioning from metered to event-based billing requires leveraging Kubernetes controllers more extensively.
+Our existing AppCat controller already handles event forwarding and webhooks, making it a natural candidate for integrating billing logic.
+
+Pros::
+
+* Full control over the lifecycle of VSHN custom resource services.
+* Customizable retry logic.
+* Clear separation of concerns between billing and service management.
+* Flexible support for different persistence backends.
+
+
+=== Runtime Library via Crossplane Composition Functions
+
+We can embed billing logic into our existing runtime library for Crossplane Composition Functions, thereby coupling service lifecycle events directly with billing logic.
+
+Pros::
+
+* Reconciliation happens directly during create, update, or delete operations.
+* Billing logic is treated as a first-class part of provisioning/deprovisioning.
+
+Cons::
+
+* Persistence integration becomes more complex.
+* No clear separation between billing and service logic.
+* Unclear where billing actually occurs, which can reduce maintainability.
+* We can't actually react on deletion events. Crossplane doesn't propagate them to the functions.
+
+== Persistence Options
+
+Based on past experience, we anticipate the need to resend older data to Odoo due to potential issues either in AppCat or Odoo.
+A lightweight, reliable mechanism to store and replay billing data is essential.
+
+Required capabilities:
+
+* Filtering and querying historical events.
+* Manual replay.
+* Partial delivery of historical events.
+* Operational simplicity
+
+=== SQLite
+
+SQLite is a simple, embedded SQL database engine suitable for local persistence needs.
+
+Pros::
+
+* Minimal setup; no external infrastructure required.
+* Fast for local and sequential read/write operations.
+* Full SQL support.
+* ACID-compliant (supports WAL mode).
+* Self-contained `.db` file that's easy to handle and back up.
+* Supports pagination and filtering by retry state or timestamp.
+
+Cons::
+
+* Not suitable for concurrent writes across multiple pods.
+* Requires manual effort for backups, failover, and compaction.
+* Not distributed or highly available.
+* Lacks integration with Kubernetes tools like `kubectl`.
+* Not inherently event-driven.
+
+=== Custom Kubernetes CRDs
+
+Custom Resource Definitions (CRDs) can be used to model billing events as native Kubernetes objects.
+
+Pros::
+
+* Native integration with Kubernetes and observable via `kubectl`.
+* Supports event-driven architectures through controllers.
+* State tracking via `.status` fields.
+* Reusable by other tools/controllers within the cluster.
+* Scales horizontally (no single-writer limitation).
+
+Cons::
+
+* Excessive CR volume may cause etcd bloat, impacting cluster performance.
+* Increased API server traffic.
+* Requires boilerplate for CRD definitions and status handling.
+* No native support for complex queries (unlike SQL).
+* Manual schema migration is necessary.
+* No built-in audit trail beyond resource versioning.
+
+== Decision
+
+=== Use Controller + Custom Kubernetes CRD
+
+The recommendation is to extend the existing AppCat controller to manage event-based billing and using Kubernetes CRDs as the persistence mechanism.
+
+*Justification:*
+
+* A controller is the natural place for billing, as it sits adjacent to service lifecycle management without coupling to it.
+* CRDs integrate well into our Kubernetes-native toolset and align with GitOps principles.
+* Data inspection and interaction via `kubectl` is simple and consistent with existing workflows.
+* While CRs are harder to query than SQL databases, we can mitigate this by providing predefined `kubectl` query templates for common tasks.
+* Kubernetes retry mechanisms can be leveraged for automatic re-delivery of failed events.
+* By using `patch` operations on CRs, we can flag specific events for manual resending to Odoo.
+* With careful CRD schema design (example: using one CR per service instead of one per event), we can avoid overwhelming etcd.
+* If detailed auditing is needed, it can be delegated to an external logging or database system.
+
+This hybrid approach gives us robust control, observability, and operational flexibility for event-based billing with minimal compromise.
+
+=== Billing Custom Resource (CR)
+
+Each Billing Custom Resource (CR) describes a single service instance and its full lifecycle - from creation to deletion.
+
+It consists of two main sections:
+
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
```diff
@@ -0,0 +1,145 @@
...
+:adr_upd_date:  2025-06-02
+:adr_status:    draft
+:adr_tags:      billing,odoo
+
+include::partial$adr-meta.adoc[]
+
+[NOTE]
+.Summary
+====
+With the upcoming support for event-based billing, our current approach for sending usage data for VSHN services needs to evolve.
```
tobru, line 15:
```suggestion
For sending billing events to Odoo (https://docs.central.vshn.ch/event-billing.html[Event Based Billing^]), well use the existing AppCat controller to manage billing events and Kubernetes CRDs as the persistence mechanism.
```

The summary in an ADR summarizes the decision.

docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
```diff
@@ -0,0 +1,145 @@
...
+====
+
+== Requirements
+
+. Reuse existing tools and/or components of AppCat to build a resilient event-based billing solution.
+. Enable resending data to Odoo (historical record retention is required).
+. Persist the state of each billing request.
+. Do not block service creation, deletion, or update due to errors in the billing system.
+. [[odoo-sync-state]] Provide a way to verify sync state between Odoo and the Kubernetes cluster.
```
tobru, line 24:
```suggestion
. Reuse existing tools and/or components of AppCat to build a resilient event-based billing solution.
. Enable resending data to Odoo (historical record retention is required).
. Persist the state of each billing request.
. Do not block service creation, deletion, or update due to errors in the billing system.
. [[odoo-sync-state]] Provide a way to verify sync state between Odoo and the Kubernetes cluster.
. Do not lose any event (created, deleted, scaled), as this has a direct financial impact.
```

tobru (changes requested):
Please structure the ADR document as described in https://kb.vshn.ch/app-catalog/adr/0001-architecture-decision-records.html

Other than that, I like the proposal and I think that could work. I'm a bit unsure about storing the events in CRDs because of the volume, maybe you want to test that a bit more and see where the limitations are.

docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
```diff
@@ -0,0 +1,145 @@
...
+
+[NOTE]
+.xref:odoo-sync-state[Odoo Sync State]
+====
+Currently, we have no reliable way to automatically verify whether the state in Odoo matches what exists in the cluster (whether stored in SQLite or as Custom Resources).
+
+At present, Odoo does not expose a `GET` endpoint to retrieve billing records, which limits our ability to perform data reconciliation.
+
+This endpoint has already been requested to improve the robustness of the system and ensure consistency between Odoo and the cluster.
+====
```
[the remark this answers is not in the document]
zugao: ok, we will discuss that within the team

docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
```diff
@@ -0,0 +1,145 @@
...
+
+The recommendation is to extend the existing AppCat controller to manage event-based billing and using Kubernetes CRDs as the persistence mechanism.
+
+*Justification:*
+
+* A controller is the natural place for billing, as it sits adjacent to service lifecycle management without coupling to it.
+* CRDs integrate well into our Kubernetes-native toolset and align with GitOps principles.
+* Data inspection and interaction via `kubectl` is simple and consistent with existing workflows.
+* While CRs are harder to query than SQL databases, we can mitigate this by providing predefined `kubectl` query templates for common tasks.
+* Kubernetes retry mechanisms can be leveraged for automatic re-delivery of failed events.
```
TheBigLee, line 130: We need to be careful here. Events are only present on a cluster for up to 1 hour. Therefore, we can't re-deliver anything older than 1 hour. This is especially true if the CR is deleted as we can't get any information from there either (eg. status field)

docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
```diff
@@ -0,0 +1,145 @@
...
+*Cons:*
+
+* Excessive CR volume may cause etcd bloat, impacting cluster performance.
+* Increased API server traffic.
+* Requires boilerplate for CRD definitions and status handling.
+* No native support for complex queries (unlike SQL).
+* Manual schema migration is necessary.
+* No built-in audit trail beyond resource versioning.
+
+== Recommendation
```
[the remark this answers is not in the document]
Kidswiss: You can still call it a decision, that's how we always did it.

docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
```diff
@@ -0,0 +1,145 @@
...
+*Justification:*
+
+* A controller is the natural place for billing, as it sits adjacent to service lifecycle management without coupling to it.
+* CRDs integrate well into our Kubernetes-native toolset and align with GitOps principles.
+* Data inspection and interaction via `kubectl` is simple and consistent with existing workflows.
+* While CRs are harder to query than SQL databases, we can mitigate this by providing predefined `kubectl` query templates for common tasks.
+* Kubernetes retry mechanisms can be leveraged for automatic re-delivery of failed events.
+* By using `patch` operations on CRs, we can flag specific events for manual resending to Odoo.
+* With careful CRD schema design (e.g., using one CR per service instead of one per event), we can avoid overwhelming etcd.
+* If detailed auditing is needed, it can be delegated to an external logging or database system.
```
Kidswiss, line 133: True, if we ever need more advanced queries we can just schlurp everything into a sql database then.

docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
Kidswiss: We should also definitely set finalizers on those objects. So we can be sure they don't get deleted by accident.

tobru:
@zugao Mind adding a bit more context to the ADR? I think an example of a creation and a deletion event in the "Context" section would help to set the stage on what events we have to send to Odoo for a better understanding.

Also an important detail: The `instance_id` has to be globally unique.

docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
zugao: yes we need

zugao:
> @zugao Mind adding a bit more context to the ADR? I think an example of a creation and a deletion event in the "Context" section would help to set the stage on what events we have to send to Odoo for a better understanding.

I added a complete example and the reasoning behind the custom resource in the Decision section.

docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
zugao: Added a finaliser in my CR example

docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
```diff
@@ -0,0 +1,233 @@
...
+*Pros:*
+
+* Full control over the lifecycle of VSHN custom resource services.
+* Customizable retry logic.
+* Clear separation of concerns between billing and service management.
+* Flexible support for different persistence backends.
+
+*Cons:*
+
+* More difficult to block or revert service operations when billing events fail to post to Odoo.
```
tobru, line 49: Why is that a con? In requirements we write that we don't want that anyway.

docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
```diff
@@ -0,0 +1,233 @@
...
+[NOTE]
+.Summary
+====
+For sending billing events to Odoo (https://docs.central.vshn.ch/event-billing.html[Event Based Billing^]), will use the existing AppCat controller to manage billing events and Kubernetes CRDs as the persistence mechanism.
+====
+
+== Requirements
+
+. Reuse existing tools and/or components of AppCat to build a resilient event-based billing solution.
+. Enable resending data to Odoo (historical record retention is required).
```
tobru, line 21:
```suggestion
. Make data resending to Odoo possible in case it's required (historical record retention is required).
```

docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
```diff
@@ -0,0 +1,233 @@
...
+* More difficult to block or revert service operations when billing events fail to post to Odoo.
+
+=== Runtime Library via Crossplane Composition Functions
+
+We can embed billing logic into our existing runtime library for Crossplane Composition Functions, thereby coupling service lifecycle events directly with billing logic.
+
+*Pros:*
+
+* Reconciliation happens directly during create, update, or delete operations.
+* Greater control over services when billing fails.
```
tobru, line 58: This actually is a con when compared to the requirements?

docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+. <<AppCat Controller>>
+. <<Runtime Library via Crossplane Composition Functions>>
+
+=== AppCat Controller
+
+Transitioning from metered to event-based billing requires leveraging Kubernetes controllers more extensively.
+Our existing AppCat controller already handles event forwarding and webhooks, making it a natural candidate for integrating billing logic.
+
+*Pros:*
```
tobru, line 40:
```suggestion
Pros::
```
Please use that format on the other pro/con listings as well. See https://docs.asciidoctor.org/asciidoc/latest/lists/description/

docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+* Minimal setup; no external infrastructure required.
+* Fast for local and sequential read/write operations.
+* Full SQL support.
+* ACID-compliant (supports WAL mode).
+* Self-contained `.db` file that's easy to handle and back up.
+* Supports pagination and filtering by retry state or timestamp.
+
+*Cons:*
```
tobru, line 92: While I agree with the cons, they barely reflect the requirements defined. But fine by me to leave it like this.

docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
```diff
@@ -0,0 +1,233 @@
...
+      type: Synced
+    - lastTransitionTime: "2023-05-25T18:45:38Z"
+      reason: Available
+      status: "True"
+      type: Ready
+----
+
+<1> An on-demand trigger to resend events from the `status.events` list based on their `state`.
+<2> Unique name of the composite - serves as the identifier for the Billing CR.
+<3> All Billing CRs reside in the `syn-appcat` namespace.
```
tobru, line 215: A rationale why this namespace would help in understanding it.

docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
```diff
@@ -0,0 +1,233 @@
...
+<4> A finalizer from the controller to protect from accidental deletion.
+<5> The field defines after how many days the CR should be deleted.
```
tobru, line 217: What happens when an instance is still running after this time has passed?

tobru (changes requested):
A few nits here and there and a few questions, but all in all, I quite like it and I think we can go this way.

docs/modules/ROOT/pages/adr/0033-event-based-billing-oddo.adoc
zugao: Indeed you are right, I will remove this. I followed general pros and cons without considering the requirements.

tobru (approved):

tobru:
In the implementation phase, we also have to figure out where to get the information for `spec.odoo` and `status.events` from. Maybe we should update the ADR then when we know that.

We do store some of the required data as labels and annotations in the organization namespace on the control-plane via the Servala Portal.

mdnix (approved):

PR #145, opened 2025-06-05
Add ADR for Redis HA

TheBigLee:
## Summary

This ADR outlines various options for running Redis in HA.

## Checklist


TheBigLee:
> Did you research how other cloud native Redis offering are doing it? Are there any Redis operators out there where we can check how they're doing it? Or an official Redis documentation, Helm Chart, white paper, architecture document, ...

I haven't researched other offerings.
There are various operators and most of them (like the helm chart) support running Redis in a Master/Replica mode or clustered mode. Both options are also documented on the official Redis documentation.

We can basically run redis either as a cluster or with Sentinel in HA/Replica setup. Those are the two officially supported ways.

docs/modules/ROOT/pages/adr/0034-vshnredis-high-availability.adoc
```diff
@@ -0,0 +1,83 @@
...
+Requirements to the Redis HA architecture:
+
+* Minimizes downtime during failover
+* Supports Kubernetes-native patterns
+* Is operationally manageable
+* Is compatible with our existing infrastructure and client usage patterns
+
+=== Considered Options
+
+**Option 1:  Redis Sentinel + Redis Master/Replica**
```
tobru, line 29:
```suggestion
Option 1:  Redis Sentinel + Redis Master/Replica::
```

Use this syntax throughout the document, ref: https://docs.asciidoctor.org/asciidoc/latest/lists/description/

docs/modules/ROOT/pages/adr/0034-vshnredis-high-availability.adoc
```diff
@@ -0,0 +1,83 @@
...
+
+Option 2: Redis Cluster Mode is too complex for our use case at the moment as it requires multiple masters and replicas.
+
+We therefore go with Option 1 and 3. They are simpler alternatives that can be implemented more easily.
+We can even use both options at the same time by using a flag if we want to enable a proxy in front of Redis.
+
+This gives us the flexibility to support users that have sentinel support in their application as well as the users that don't.
+Furthermore, with Option 3 we ensure that the service is also usable with Servala, where the service is exposed publicly.
+
+Last but not least, Options 1 and 3 make it also easy to migrate from HA to non-HA and vice versa. With Option 2 this is not easily doable, as the data is spread across different masters.
```
tobru, line 78: The first part of this section should contain the decision. When I started to read I thought we decided on option 2. Please see https://kb.vshn.ch/app-catalog/adr/0001-architecture-decision-records.html how to structure an ADR document.

TheBigLee:
> Did you research how other cloud native Redis offering are doing it? 
Yes I did some research and all the big companies (Amazon, Google, MS, Heroku, Digital Ocean) use the Master/Replica approach and not the Cluster approach and all of them seem to use some kind of proxy as there is no mention of sentinel in any of the products I've checked.

> Are there any Redis operators out there where we can check how they're doing it? 
I haven't looked into any operators since this is not the topic of this ADR. Our helm chart already supports all officially supported running modes of Redis and so do most operators. Switching from a helm based setup to an operator is a lot of work and out of scope for this ADR.

> Or an official Redis documentation, Helm Chart, white paper, architecture document, ...
I've read through the official Redis documentation and various blogs that demonstrate how to run Redis in HA.
The official documentation and also most blogs document two ways of running Redis in HA and both have been described in this ADR (Master/Replica & Clustering)

TheBigLee:
> Speaking of the loadbalancer, there's https://github.com/twitter/twemproxy which seems to be more native for redis than haproxy/envoy. Might be worth taking a look.

This proxy seems dead. Last commit was in October 2022

wejdross (commented):
I think I can see 4th option:
let's write a very small logic to just keep watching redis and sentinels and detect master switches and update pods labels accordingly

This way we could use native k8s service and ingresses, expose master and replicas under one additional, paid ipv4

We could deploy that small "controller" per namespace anmd scrape redis endpoints to detect when it's switched, like 1 second interval and then if changed update pod label with master, service picks new master, ingress routes trraffic to new master immediately. No external proxies, no complex setups, bulletproof logic

TheBigLee:
> I think I can see 4th option: let's write a very small logic to just keep watching redis and sentinels and detect master switches and update pods labels accordingly
> 
> This way we could use native k8s service and ingresses, expose master and replicas under one additional, paid ipv4
> 
> We could deploy that small "controller" per namespace anmd scrape redis endpoints to detect when it's switched, like 1 second interval and then if changed update pod label with master, service picks new master, ingress routes trraffic to new master immediately. No external proxies, no complex setups, bulletproof logic

I like this idea. I'll draft it up properly and will add it do the ADR

docs/modules/ROOT/pages/adr/0034-vshnredis-high-availability.adoc
```diff
@@ -0,0 +1,100 @@
...
+  ** No need for Sentinel
+  ** Client remains unaware of master failover; proxy handles redirection
+  ** Greater flexibility in failover and health checks
+  ** Can be exposed outside of Kubernetes (for example Loadbalancer service) and therefore be used by Servala
+* Cons:
+  ** Requires custom failover scripts or integration with Sentinel to update routes
+  ** Slight increase in latency due to proxy hop
+  ** Operational complexity: proxy config must track Redis role changes
+
+Option 4: Redis Master/Replica with Label-based Service Selector::
```
Kidswiss, line 68: Did you do a test to see how long K8s takes to actually change the traffic to another pod?

For patroni it takes about 10-15 seconds. Patroni does it similarly, but there's also a lot of PostgreSQL failover shenanigans going on. So it might even be faster for Redis, but probably slower than using a proper proxy.

TheBigLee: It's actually pretty fast. Here is a screencast of this in action:

https://github.com/user-attachments/assets/9301abc7-5b73-4a47-b7c3-6001993fd6a1

TheBigLee:
I did some more rewording. Techically option 4 is not a controller, but a regular deployment that updates labels on a pod. It doesn't reconcile against any resources but rather watches for redis master changes on the sentinel port.

zugao (commented):
As already discussed face-to-face,, option 4 seems very appealing. Instead of using a controller or an operator we can directly use `master.sidecar` parameter from the Helm Chart and inject it in our comp function a script that would check on master change and would label the pod accordingly. This solution seems minimal and arguably quite easy to implement but, I have no experience with using Redis and I don't know if this would be enough to substitute a full fledged HAProxy.

zugao (approved):
I cautiously approve, but I have the same worries as @mdnix. I would like to avoid a bottleneck with this solution if this does not work properly. We could end up spending too much time with it. Nevertheless I still believe we should try it out as I don't expect a difficult implementation.

Update: Checking a bit around I noticed that we will have to manage edge cases in our script/go app such as:
* TCP connection draining/retries
* Redis-specific edge cases (e.g., pipeline, pub/sub, etc.)
* Crash recovery

I am more inclined to just go with HAProxy especially when we expect to have customers with big Redis instances.

mdnix (approved):
Personally, I’m in favor of option 4, even if it requires some additional engineering effort.
If we decide not to go that route my next preference would be the HAProxy (or some other proxy) based solution.

Kidswiss:
I'm also in favor of Option 4.

This way we don't have any additional stuff in the datapath between Redis and the client.

PR #194, opened 2026-04-21
Add ADR for Standardized status fields

TheBigLee:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
+++ b/docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
@@ -185,7 +185,8 @@ type VSHNPostgreSQLStatus struct {
 | `instanceNamespace` not yet set OR initial resource creation
 
 | `Ready`
-| Crossplane `Ready=True`, `Synced=True`, not in maintenance
+| Crossplane `Ready=True`, `Synced=True`, all service pods in `instanceNamespace` are `Ready`, not in maintenance.
+See <<pod-readiness>> below.
                                                                                                                                                                                                             
 | `Maintenance`
 | Maintenance window active (tracked by maintenance controller)                                                                                                                                             
@@ -211,6 +212,54 @@ Priority when multiple conditions active:
 ----
 Deleting > Failed > Provisioning > Maintenance > Scaling > Degraded > Ready > Unknown
 ----
+
+[#pod-readiness]
+==== Pod Readiness
+
+All AppCat services are deployed via Helm. Crossplane marks a service `Ready=True` and `Synced=True` as soon as the Helm release is successfully applied.
+It does not track pod lifecycle. This means a service can report `Ready` while its pods are still initializing,
+terminating from a previous revision, or in `CrashLoopBackOff`.
+
+To guard against premature `Ready` phase the composition function MUST perform an additional pod readiness check:
+
+. After Crossplane conditions are `Ready=True` and `Synced=True`, query all pods in `instanceNamespace`.
+. If any pod is not in `Running` phase with all containers `Ready`, remain in `Provisioning` (initial deploy) or
+  transition to `Degraded` (post-deploy regression) rather than setting `phase=Ready`.
+. Only set `phase=Ready` when Crossplane conditions AND pod readiness are both satisfied.
+
+This check is implemented inside the composition function (function-appcat). It sets a custom condition
+`PodReady` (type `PodReady`, status `True`/`False`) on the composite resource alongside the standard
+Crossplane conditions, giving operators visibility into which layer is not yet healthy:
+
+[cols="2,1,1,2",options="header"]
+|===
+| Scenario | Crossplane `Ready` | `PodReady` condition | `phase`
+
+| Initial provisioning, pods pending
+| False
+| False
+| `Provisioning`
+
+| Helm release applied, pods not yet ready
+| True
+| False
+| `Provisioning`
+
+| All pods running and ready
+| True
+| True
+| `Ready`
+
+| Post-update pod regression (was ready before)
+| True
+| False
+| `Degraded`
+
+| Helm release failed
+| False
+| False
+| `Failed`
+|===
 
 === Internal Conditions Cleanup                                                                                                                                                                             
 
--- a/docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
+++ b/docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
@@ -1,4 +1,4 @@
-= ADR 0051 - Standarized Status Fields for AppCat Services
+= ADR 0051 - Standardized Status Fields for AppCat Services
 :adr_author:    Nicolas Bigler
 :adr_owner:     Schedar
 :adr_reviewers: 
@@ -12,9 +12,7 @@ include::partial$adr-meta.adoc[]
 [NOTE]
 .Summary
 ====
-Introduce a shared `VSHNServiceStatus` struct embedded in all AppCat service status types. It exposes a standardized
-`phase` enum (`Provisioning`, `Ready`, `Scaling`, `Maintenance`, `Degraded`, `Failed`, `Deleting`, `Unknown`) with 
-optional `phaseReason` and `phaseMessage` fields. This gives consumers (for example Servala portal) a single machine-readable lifecycle field without service-specific knowledge, while keeping existing condition-based detail for operators.
+Introduce a shared `VSHNServiceStatus` struct embedded in all AppCat service status types using a hybrid approach. It exposes a `phase` enum (`Provisioning`, `Ready`, `Scaling`, `Maintenance`, `Degraded`, `Failed`, `Deleting`, `Unknown`) for quick machine-readable summary, plus standard `metav1.Condition[]` for rich transition history and detail. This aligns with the Crossplane way of working: consumers can read `phase` for a simple lifecycle view while operators and automation can inspect conditions for full history and reasoning.
 ====
 
 == Context
@@ -121,13 +119,13 @@ No new fields. Rely entirely on existing `xpv1.ResourceStatus` conditions
 
 == Decision
 
-Introduce a standard `VSHNServiceStatus` struct embedded in all service status types.                                                                                                                       
-  
-=== Standard Fields (all services)                                                                                                                                                                          
-                                                                                                                                              
+Use the hybrid approach (Option 3): a single `VSHNServiceStatus` struct with a `phase` enum for quick summary and `metav1.Condition[]` for history and detail. No intermediate wrapper struct.
+
+=== Standard Fields (all services)
+
 [source,go]
 ----
-type VSHNServiceStateStatus struct {
+type VSHNServiceStatus struct {
     // Phase summarizes the lifecycle state of the service.
     // +kubebuilder:validation:Enum=Provisioning;Ready;Scaling;Maintenance;Degraded;Failed;Deleting;Unknown
     // +kubebuilder:default=Unknown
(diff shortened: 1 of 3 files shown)


docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
```diff
@@ -0,0 +1,250 @@
...
+:adr_upd_date:  2026-04-20
+:adr_status:    draft
+:adr_tags:      framework,servala,service
+
+include::partial$adr-meta.adoc[]
+
+[NOTE]
+.Summary
+====
+Short and concise summary of the decision in this ADR.
```
mikeshootzz: Can we make this more of a summary? :D

mikeshootzz (approved):
Rest LGTM

docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
TheBigLee: Right ;-) Fixed

zugao (commented):
I expected a wider research footprint then just setting specific status fields. For instance we don't have here billing information which should be added in the future (There is a topic for that in Framework 2.0). The solution should be easily extensible so the scope isn't just the current state of a service but wider. Now I am not sure if we want to tackle the problem in this ADR but it would be necessary to have a clearer picture what kind of service status we want to have. Also what about service specific status fields? where those have to be made avaialble?

Also the `Scaling detection` question can be solved with Crossplane 2.x naturally using a separate comp function.

zugao (commented):
I liked an idea here where we have a separate item where we save the current state of the service, unlike the current k8s approach where you have to search in the list which status is active. I would discuss this with other engineers to see if it's viable solution. We can have an item outside the array and the rest inside the array (history)

TheBigLee:
> I liked an idea here where we have a separate item where we save the current state of the service, unlike the current k8s approach where you have to search in the list which status is active. I would discuss this with other engineers to see if it's viable solution. We can have an item outside the array and the rest inside the array (history)

I've moved the `Phase` fields to a sub struct like we do in the billing example. This should be more consistent and easier to extend

TheBigLee:
Let's discuss the details first in the next tech talk, before going further with the ADR

docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
```diff
@@ -0,0 +1,292 @@
...
+
+[cols="1,3",options="header"]                                                                                                                                                                               
+|===
+| Phase | Condition                                                                                                                                                                                         
+                                                                                                                                              
+| `Provisioning`
+| `instanceNamespace` not yet set OR initial resource creation
+
+| `Ready`
+| Crossplane `Ready=True`, `Synced=True`, not in maintenance
```
Kidswiss: One complaint I heard a lot over the time: "It says ready, but the container is still starting!?"

If we want to rely on the crossplane readiness, we should ensure to have proper readiness checks in the comp-functions and only make something ready if the pods are actually running.

docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
```diff
@@ -0,0 +1,292 @@
...
+=== Negative
+
+* Composition functions must set `phase` correctly. Ongoing maintenance burden
+* Phase transitions need to be tested
+* Scaling detection requires comparing spec to actual state (non-trivial)
+
+== Open Questions
+
+. *Servala specifics:* Do we need additional states beyond the 8 proposed?
+. *Scaling detection:* How to reliably detect scaling in progress without polling?
```
Kidswiss: My 2 cents: we make sure to set things unready when a scaling operation is triggered and then only set it ready again once we see the expected state. That's how we've solved these kinds of issues in the past. It can get very fiddly though...

We can also explore Crossplane 2.x's realtime/watch functionality.

Kidswiss (commented):
I like the hybrid approach. Crossplane providers work somewhat similar.

They don't use conditions but events instead. But the idea is the same:

They throw various events depending on the state of the CR. If there's an error then the relevant event is added to the unready condition. The only downside of events: they are ephemeral, so if you find a broken instance in the morning and all events have been purged, you won't have any historic data anymore.

By using conditions and a dedicated field we have better insights about the various phases a composite/claim went through and are able to detect transient failures.

docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
```diff
@@ -0,0 +1,292 @@
+= ADR 0051 - Standarized Status Fields for AppCat Services
```
mdnix: Should be "Standardized". This appears multiple times across this PR

docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
```diff
@@ -0,0 +1,292 @@
...
+- No unified "phase" or "state" field across services
+- Transition states (provisioning, scaling, maintenance) are not exposed
+- Some services expose internal implementation details
+- Inconsistent depth (PostgreSQL has 11 conditions, Garage has none)
+
+=== Solution Ideas
+
+==== Option 1: Single phase field (enum string)
+
+Add a top-level `phase` field (for example `Provisioning`, `Ready`, `Scaling`, `Maintenance`, `Degrated`, `Failed`,
```
mdnix:
```suggestion
Add a top-level `phase` field (for example `Provisioning`, `Ready`, `Scaling`, `Maintenance`, `Degraded`, `Failed`, 
```

docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
```diff
@@ -0,0 +1,292 @@
...
+
+== Decision
+
+Introduce a standard `VSHNServiceStatus` struct embedded in all service status types.                                                                                                                       
+  
+=== Standard Fields (all services)                                                                                                                                                                          
+                                                                                                                                              
+[source,go]
+----
+type VSHNServiceStateStatus struct {
```
mdnix: Maybe something like `VSHNServicePhaseInfo` or similar instead?

docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
```diff
@@ -0,0 +1,292 @@
...
+| No migration needed                                                                                                                                                                                       
+| Consumers must understand Crossplane internals                                                                                                                                                            
+                                                                                                                                                                                                            
+| Consistent with upstream conventions                                                                                                        
+| `Synced=False` covers too many distinct failure modes
+|===
+
+== Decision
+
+Introduce a standard `VSHNServiceStatus` struct embedded in all service status types.
```
mdnix: Just for clarity, there should be a sentence like "We chose option xyz..."

docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
```diff
@@ -0,0 +1,292 @@
...
+
+Introduce a standard `VSHNServiceStatus` struct embedded in all service status types.                                                                                                                       
+  
+=== Standard Fields (all services)                                                                                                                                                                          
+                                                                                                                                              
+[source,go]
+----
+type VSHNServiceStateStatus struct {
+    // Phase summarizes the lifecycle state of the service.
+    // +kubebuilder:validation:Enum=Provisioning;Ready;Scaling;Maintenance;Degraded;Failed;Deleting;Unknown
```
mdnix, line 130: Do we need `Unknown` here is default already falls back to it?

mdnix (commented):
Maybe we could get inspired from upstream Kubernetes a bit here. I checked a bit how they do it with `PodStatus`, `NodeStatus` and so on. https://github.com/kubernetes/api/blob/master/core/v1/types.go#L5307

That would look a little like this for our use case

```go
type VSHNServicePhase string

const (
    PhaseProvisioning VSHNServicePhase = "Provisioning"
    PhaseReady        VSHNServicePhase = "Ready"
    PhaseScaling      VSHNServicePhase = "Scaling"
    PhaseMaintenance  VSHNServicePhase = "Maintenance"
    PhaseDegraded     VSHNServicePhase = "Degraded"
    PhaseFailed       VSHNServicePhase = "Failed"
    PhaseDeleting     VSHNServicePhase = "Deleting"
    PhaseUnknown      VSHNServicePhase = "Unknown"
)

type VSHNServiceStatus struct {
    Phase              VSHNServicePhase `json:"phase,omitempty"`
    Reason             string           `json:"reason,omitempty"`
    Message            string           `json:"message,omitempty"`
    LastTransitionTime *metav1.Time     `json:"lastTransitionTime,omitempty"`
}

// Embed the status inline
type VSHNPostgreSQLStatus struct {
    VSHNServiceStatus `json:",inline"`
    // ...existing fields...
}
```

docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
TheBigLee: :+1: Will update it once we decided on an option

TheBigLee:
As discussed: We will do a hybrid approach with conditions and fields, this has the benefit of being easily parsable by servala and also gives us a status history through the conditions array

docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
TheBigLee: Fixed

docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
TheBigLee: We actually need it in both as far as I know, or the default "Unknown" state would not be accepted as it would be missing in the allowed phases in the OpenAPI spec

docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
TheBigLee: I've added a section for that scenario as well.

docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
TheBigLee: I've adjusted the ADR to reflect this

docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
TheBigLee: I've removed the middle layer struct, so this is obsolete

docs/modules/ROOT/pages/adr/0051-standarized-status-fields-for-appcat-services.adoc
```diff
@@ -0,0 +1,341 @@
...
+==== Pod Readiness
+
+All AppCat services are deployed via Helm. Crossplane marks a service `Ready=True` and `Synced=True` as soon as the Helm release is successfully applied.
+It does not track pod lifecycle. This means a service can report `Ready` while its pods are still initializing,
+terminating from a previous revision, or in `CrashLoopBackOff`.
+
+To guard against premature `Ready` phase the composition function MUST perform an additional pod readiness check:
+
+. After Crossplane conditions are `Ready=True` and `Synced=True`, query all pods in `instanceNamespace`.
+. If any pod is not in `Running` phase with all containers `Ready`, remain in `Provisioning` (initial deploy) or
```
Kidswiss:
```suggestion
. If any service pod is not in `Running` phase with all containers `Ready`, remain in `Provisioning` (initial deploy) or
```

Namespaces can contain pods that are not relevant to the service's readiness, for example backup or maintenance pods, which can be in completed phases.

Kidswiss (changes requested):

mikeshootzz (approved):

zugao (approved):
LGTM

mdnix (approved):

Kidswiss (approved):

PR #195, opened 2026-04-24
Add ADR for day 2 operations with AppCat

mikeshootzz:
## Summary

This defines how we handle day 2 operations for AppCat moving forward.

## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
+++ b/docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
@@ -3,7 +3,7 @@
 :adr_owner:     Schedar
 :adr_reviewers: Schedar
 :adr_date:      2026-04-23
-:adr_upd_date:  2026-04-27
+:adr_upd_date:  2026-04-30
 :adr_status:    draft
 :adr_tags:      framework,framework2,operations,maintenance,crossplane,helm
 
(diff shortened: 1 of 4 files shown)


docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
```diff
@@ -0,0 +1,213 @@
...
+[NOTE]
+.Summary
+====
+This ADR defines where day 2 operations logic lives in AppCat services and which mechanism implements each class of operation.
+
+Business logic that shapes Kubernetes objects belongs in the service Helm chart.
+Composition functions return desired state only.
+One-off operations are Jobs emitted by the composition function, running `appcat operation <op>` (mirroring the maintenance pattern); Crossplane Operations remain a long-term target.
+Recurring operations use Kubernetes Jobs driven by CronJobs.
+Observation of existing cluster state uses https://docs.crossplane.io/latest/composition/compositions/#required-resources[Crossplane required resources].
```
Kidswiss: We could probably refactor parts of the runtime to make this change seamless.

docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
```diff
@@ -0,0 +1,213 @@
...
+* Bumping `function-sdk-go` to expose the new `required_resources` proto field cascades into `vshn/apiserver-runtime` fork breakage.
+* Operations cannot delete resources; no built-in cleanup story.
+* Failure feedback lands on the `Operation` CR, not on the originating service claim; surfacing requires extra plumbing.
+
+==== Option B: Job emitted by the composition function
+
+Description::
+
+The composition function emits a Job whose container runs `appcat operation <subcommand>`.
+A shared helper in `pkg/comp-functions/functions/common` builds the Job, mirroring the maintenance helper.
```
Kidswiss: I'd say `pkg/operations` would make more sense.

docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+Gating uses stable, identity-keyed Job names so re-emission is a no-op once the Job exists. Two patterns:
```
Kidswiss: We'd need some way to share data back with the comp-function.

docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+* Reuses the maintenance pattern already in production.
+* Operation runtime lives in the `appcat` binary, alongside maintenance logic.
+* Job lifecycle, logging, status, and cleanup are well-understood Kubernetes primitives.
+* No alpha dependencies; no extra Crossplane plumbing.
+* Failure feedback lands on a Job that the composition function already manages.
+
+Disadvantages::
+
+* Operation logic written in Go inside the `appcat` binary adds to that binary's responsibilities.
```
Kidswiss: If we need to manipulate K8s objects via the operation jobs, we also need to handle RBAC.

docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
```diff
@@ -0,0 +1,213 @@
...
+| **Chosen**
+| Native Crossplane primitive with explicit read-only ownership semantics.
+
+| B
+| `SetDesiredKubeObject` + `KubeOptionObserve` (read back via `GetObservedKubeObject`)
+| Existing pattern, not refactored
+| Widely used in the codebase today; less explicit about read-only intent.
+|===
+
+New implementations declare observed-but-unowned resources as https://docs.crossplane.io/latest/composition/compositions/#required-resources[required resources]. Existing `KubeOptionObserve` usage is not refactored as part of this ADR.
```
Kidswiss: > usage is not refactored as part of this ADR.

Why not though? The runtime will be harder to use if there's two methods to do the same thing. We already have the issue that we have multiple implementations for the same things which make the comp-functions very inconsistent.

docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
mikeshootzz: That's a very valid point :D

docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
mikeshootzz: Fair

docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
mikeshootzz: That's true. The idea behind this was that refactoring takes a good chunk of time and wanting to move forward with 2.0 but you're right, it makes things worse

docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
```diff
@@ -0,0 +1,214 @@
...
+| Widely used in the codebase today; less explicit about read-only intent.
+|===
+
+New implementations declare observed-but-unowned resources as https://docs.crossplane.io/latest/composition/compositions/#required-resources[required resources]. Existing `KubeOptionObserve` usage will be migrated to required resources to avoid two parallel mechanisms in the codebase.
+
+== Decision
+
+=== Composition functions return desired state only
+
+Composition functions describe steady state for a given claim. They do not perform one-off actions, mutate external systems, or omit resources conditionally as a form of imperative control.
```
zugao: What is `steady state` ?

> They do not perform one-off actions

Aren't Jobs in composition functions (Option B) a one off action?

> omit resources conditionally as a form of imperative control.

If I understood this correctly, we do it all the time in composition functions, we remove, we add resources based on various conditions at runtime. I know we want to move almost all the abstraction into the helm chart but have we investigated all the cases in our existing services, is it achievable? Also what do we do with upstream helm charts? Do we fork and managed the charts ourselves?

docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
```diff
@@ -0,0 +1,214 @@
...
+
+If a piece of logic cannot be expressed as "given this input, this is the desired resource," it does not belong in the composition function.
```
zugao: I don't understand the logic of this sentence. The input is the claim, the desired state is defined by composition functions and will have a desired state how we want it to be.

docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
```diff
@@ -0,0 +1,214 @@
...
+
+=== Helm charts own Kubernetes object shape
+
+The service Helm chart is the single source of truth for the shape of the service's Kubernetes objects. The composition function sets chart values; it does not template workload manifests itself. Forks of upstream charts are acceptable when the upstream is insufficient.
```
zugao: > Forks of upstream charts are acceptable when the upstream is insufficient

I would argue that any third party chart we adopt will require forking. Look at the story regarding exposing services via TCP, no helm chart exposes that kind of detail. There will be a huge operational overhead to maintain the, It's just one example and there are so many that we will need to handle all helm charts ourselves. The beauty of composition function is that we can reuse code to apply the same logic across multiple service. We can do that in helm charts as well as we have access to some stuff at runtime and we might be able to reuse some templates but is it flexible enough like go?

docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
mikeshootzz: I think this needs a bit more discussion/clarification and adjustment of the wording. Can we discuss this on Thursday when we look at this?

docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
```diff
@@ -0,0 +1,214 @@
...
+
+=== One-off operations use Jobs emitted by the composition function
+
+Use Option B: the composition function emits a Job whose container runs `appcat operation <subcommand>`. New operations are added by writing the subcommand and wiring it through the shared helper in `pkg/operations`.
+
+Crossplane Operations (Option A) remain the long-term target; re-evaluation is deferred per the Phase 2 note in xref:adr/0047-service-maintenance-and-upgrades-framework-2-0.adoc[ADR 0047]. A dedicated day-2 ops chart (Option E) will be reconsidered if a single ops Job becomes shared across services.
+
+=== Recurring operations use Kubernetes Jobs
+
+Recurring work (maintenance-window upgrades, version checks, scheduled backups, periodic housekeeping) runs as Jobs scheduled by CronJobs, per xref:adr/0047-service-maintenance-and-upgrades-framework-2-0.adoc[ADR 0047]. The CronJob is built by the composition function and runs `appcat maintenance`. The service Helm chart is not involved.
```
zugao: Architecturally speaking, one time jobs should not be triggered in our composition functions, this is a suboptimal solution. But I understand that big changes should happen in Framework 2.0 (as you stated with Option A, we also hve Argo Workflows which might also be solution) and we can just keep what we are doing already

docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
zugao: absolutely, this is a critical and difficult topic, we should definitely need to discuss this together.

lieneluksika:
From refinement:
Conceptual (how we do things) vs technical decision (how we get there). 

This ADR are for 1.0 to fix the current pain points as easily as possible. 2.0 will be handled in the workshop. 
Real examples of what goes in comp functions and what in helm charts.

docs/modules/ROOT/pages/adr/0053-day-2-operations-with-appcat.adoc
```diff
@@ -0,0 +1,98 @@
...
+
+* Derives from desired values alone (no need to read current cluster state)
+* Is part of the service steady state—always present or always conditionally present based on values
+* Is best expressed as a template (labels, resource limits, conditional sections, config rendering)
+
+Example: rendering a `PostgresConfig` with user-supplied parameters, enabling or disabling a sidecar based on a feature flag, setting resource requests based on instance size.
+
+=== When logic belongs in the composition function
+
+Put logic in the composition function when it:
```
Kidswiss, line 96: I would add another point here. Although it is a special case, but I feel we need to define it, otherwise we will get headaches later on:

* It does NOT change any helm managed resources

There are multiple thoughts behind this rule:
* If a resource is managed at two different places, it gets really hard to debug
* If helm and comp-functions don't agree on the state of a resource it can cause really weird bugs

As a consequence we might want to note that using `lookup` in helm charts should be encouraged to avoid having the comp-function patch helm managed objects.

Kidswiss (commented):
Just one thing I'd like to put on the table.

docs/modules/ROOT/pages/adr/0053-day-2-operations-with-appcat.adoc
```diff
@@ -0,0 +1,98 @@
...
+
+* Needs to observe current cluster state to decide what to do
+* Represents a one-off reaction to a lifecycle event (upgrade completed, restore requested, version mismatch detected)
+* Must be reflected on the claim status (the user or external tooling needs to know it happened)
+* Has a distinct start, progress, and completion—a state machine, not a template
```
Kidswiss, line 101: I like this, we should treat the comp functions as a state machine.

docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
```diff
@@ -0,0 +1,98 @@
...
+| Suitable for
+| Object shape, configuration, service structure
+| Lifecycle events, reactive operations, progress tracking
+|===
+
+=== When logic belongs in the Helm chart
+
+Put logic in the chart when it:
+
+* Derives from desired values alone (no need to read current cluster state)
```
Kidswiss: This clashes with `lookup` though. Maybe we should re-phrase it?

zugao (commented):
We have to talk about this again. The current composition function is not meant for one time jobs or events, that's why the whole new branch of Operations exists in Crossplane. The whole problem is that the majority of our services are in cluster objects, not cloud objects, this means that what Crossplane can do, Helm does it as well. 
We can keep Crossplane for one time jobs but this will have to change rather soon with Crossplane day 2 ops...

docs/modules/ROOT/pages/adr/0053-day-2-operations-with-appcat.adoc
mikeshootzz: Adressed this

docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
mikeshootzz: fixed

mikeshootzz:
> We have to talk about this again. The current composition function is not meant for one time jobs or events, that's why the whole new branch of Operations exists in Crossplane. The whole problem is that the majority of our services are in cluster objects, not cloud objects, this means that what Crossplane can do, Helm does it as well. We can keep Crossplane for one time jobs but this will have to change rather soon with Crossplane day 2 ops...

@zugao Added a clear scope at the beginning

docs/modules/ROOT/pages/adr/0052-day-2-operations-with-appcat.adoc
```diff
@@ -0,0 +1,119 @@
...
+* xref:adr/0047-service-maintenance-and-upgrades-framework-2-0.adoc[ADR 0047] — CronJobs for scheduled maintenance
+
+== Scope
+
+This ADR applies to *Framework 1.0* — the current AppCat architecture using Crossplane composition functions and Helm charts.
+It establishes a conceptual boundary (where does logic belong?) not a technical implementation plan.
+
+In scope:
+
+* Deciding whether new reactive or lifecycle-driven operations go in the Helm chart or the composition function
```
Kidswiss: NIT, but to make it really implementation agnostic:

Instead of writing "composition functions" use "appropriate Crossplane features", this leaves if open for Crossplane's operations, cronjobs, or whatever else we could use. So people don't get fixated on this.

Kidswiss (approved):

zugao (approved):

TheBigLee (approved):

PR #198, opened 2026-05-27
ADR for managed Forgejo Runners

mikeshootzz:
## Summary

This ADR describes how we can provide managed Forgejo runners for VSHNForgejo/Codey

## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/adr/0054-vshn-managed-forgejo-runners.adoc
+++ b/docs/modules/ROOT/pages/adr/0054-vshn-managed-forgejo-runners.adoc
@@ -42,11 +42,11 @@ This leaves us with two design questions:
 The https://codeberg.org/wrenix/helm-charts[`oci://codeberg.org/wrenix/helm-charts/forgejo-runner`] chart is used to deploy the runner and was already used in the PoC.
 It exposes everything we need:
 
-* `runner.config.existingSecret` — points the runner at a pre-built `.runner` credential file, skipping the chart's own registration job.
+* `runner.config.existingSecret`: points the runner at a pre-built `.runner` credential file, skipping the chart's own registration job.
   This is what makes the `provider-http` registration approach work.
-* `dind.*` values — built-in Docker-in-Docker sidecar support.
-* `runner.config.file.container.privileged` — defaults to `false` for job containers.
-* `securityContext.privileged` — defaults to `true` (required for DinD); must be set to `false` for the rootless option.
+* `dind.*` values: built-in Docker-in-Docker sidecar support.
+* `runner.config.file.container.privileged`: defaults to `false` for job containers.
+* `securityContext.privileged`: defaults to `true` (required for DinD); must be set to `false` for the rootless option.
 
 The chart can be forked if we need to tweak it for our use case in the future, similar to how we handle other third-party charts.
 
--- a/docs/modules/ROOT/pages/adr/0054-vshn-managed-forgejo-runners.adoc
+++ b/docs/modules/ROOT/pages/adr/0054-vshn-managed-forgejo-runners.adoc
@@ -37,6 +37,22 @@ This leaves us with two design questions:
 
 == Solutions
 
+=== Helm chart
+
+The https://codeberg.org/wrenix/helm-charts[`oci://codeberg.org/wrenix/helm-charts/forgejo-runner`] chart is used to deploy the runner and was already used in the PoC.
+It exposes everything we need:
+
+* `runner.config.existingSecret` — points the runner at a pre-built `.runner` credential file, skipping the chart's own registration job.
+  This is what makes the `provider-http` registration approach work.
+* `dind.*` values — built-in Docker-in-Docker sidecar support.
+* `runner.config.file.container.privileged` — defaults to `false` for job containers.
+* `securityContext.privileged` — defaults to `true` (required for DinD); must be set to `false` for the rootless option.
+
+The chart can be forked if we need to tweak it for our use case in the future, similar to how we handle other third-party charts.
+
+NOTE: The chart maintainer notes they have switched to Woodpecker CI and may not test every change.
+This is worth keeping in mind for future upgrades.
+
 === Runner topology
 
 * *Shared pool*: a single runner pool shared across all customer instances.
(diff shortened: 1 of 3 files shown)


docs/modules/ROOT/pages/adr/0054-vshn-managed-forgejo-runners.adoc
```diff
@@ -0,0 +1,179 @@
...
+* CI jobs from one customer must not be able to access another customer's workloads or data.
+* Runner must support arbitrary CI workloads including Docker builds.
+* Runner and job containers must run unprivileged; compatible with OpenShift restricted SCCs.
+* Resource usage must be isolatable from the Forgejo workload.
+
+== Solutions
+
+=== Helm chart
+
+The https://codeberg.org/wrenix/helm-charts[`oci://codeberg.org/wrenix/helm-charts/forgejo-runner`] chart is used to deploy the runner and was already used in the PoC.
```
mdnix: There's actually this chart upstream, have you checked it out? -> https://code.forgejo.org/forgejo-helm/forgejo-runner

mikeshootzz: I didn't find that 🤦 thanks! The chart is basically a direct adoption of what I used and is structurally the same. Updated the ADR to reflect this.

zugao (commented):
Looks good, and I agree with the general architecture. However, I think we should reconsider the Runner topology and instance namespace solution.
If we classify this as an AddOn (similar to Collabora), it should reside in the forgejo instance namespace. A runner cannot be compared to a standalone service like Postgres, it cannot function without a Forgejo instance. Because it is tightly coupled and not a separate, keeping it within the same namespace makes the most sense.

zugao (approved):

docs/modules/ROOT/pages/adr/0054-vshn-managed-forgejo-runners.adoc
```diff
@@ -0,0 +1,402 @@
...
+| ✅ no extra namespace wiring
+| ⚠️ must provision and wire the dedicated namespace
+| ❌ highest: provision a VM on the CSP, install/register the runner, wire networking; depends on a CSP Crossplane provider
+|===
+
+The CSP compute option also unblocks privileged / Docker-in-Docker execution.
+On a shared cluster, a privileged job or a container escape reaches the node, and potentially other tenants and the control plane, so it must be avoided (see <<job-execution-and-isolation>>).
+On a dedicated VM, ideally ephemeral (`--ephemeral`, one job per runner), the blast radius of an escape is the disposable VM itself.
+That makes `container.privileged = true` and DinD an acceptable trade-off for builds that genuinely need them.
+The cost is provisioning latency, recurring CSP compute spend, and a dependency on a per-CSP provisioning provider.
```
tobru, line 117: > recurring CSP compute spend

We also have this when running a Pod, it's just not so directly visible

docs/modules/ROOT/pages/adr/0054-vshn-managed-forgejo-runners.adoc
```diff
@@ -0,0 +1,402 @@
...
+
+A natural concern is whether the runner spawns additional job Pods, and if so whether their number per instance must be capped, since this drives capacity management and resource billing.
+
+With the `forgejo-runner` chart it does *not* spawn extra Pods per job.
+A CI run is confined to the pre-provisioned runner Pod(s): jobs execute inside the running runner, and when all runner capacity is busy, further jobs queue until a runner frees up rather than scaling out new Pods.
+(Whether multiple jobs can share a single Pod concurrently depends on the runner's `capacity` setting; in the PoC the chart ran jobs within the provisioned Pod rather than fanning out.)
+
+This bounds resource and storage usage to the provisioned runner Pods, which keeps capacity and billing straightforward.
+Concurrency is determined by the configured runner `capacity` and the number of runner Pods per instance, both fixed at provisioning time, instead of an unbounded pool of dynamically spawned job Pods.
+Customers therefore pick from a small set of pre-defined runner sizes (CPU / memory / storage), and the instance's runner footprint is known up front.
```
tobru, line 242: Just want to emphasize that this is very important for the product and changing this has lots of effect on the productization. So we need to take care that this doesn't change in the future by pure chance without revisiting productization.

docs/modules/ROOT/pages/adr/0054-vshn-managed-forgejo-runners.adoc
```diff
@@ -0,0 +1,402 @@
...
+If a dedicated namespace later proves worthwhile (for example for the runner-only AddOn case, or for stricter resource boundaries), we can move it without revisiting the rest of this decision.
+
+Namespace isolation is sufficient on our target platforms, which is also why we prefer the in-cluster approach over CSP VMs.
+The isolation story differs per platform:
+
+* On *Servala*, we additionally isolate the runners with a hardened `RuntimeClass` (gVisor), assigned via the runner Pod's runtime class, putting a syscall-sandbox boundary around the job containers on top of the Talos-based hardening that already goes beyond vanilla Kubernetes.
+  The initial version does not add a separate node pool; we may introduce one later to separate the runners (a bursty, interruptible workload) from platform and other-customer workloads.
+* On *APPUiO Public*, no such sandboxing is needed: customers are already restricted by OpenShift SCCs and can provision arbitrary resources in their own namespaces behind quotas and existing safeguards.
+* On *APPUiO Managed*, the cluster belongs to the customer, who can deploy whatever they like on it, so runner isolation is not our concern.
```
tobru:
```suggestion
* On *APPUiO*, no such sandboxing is needed: customers are already restricted by OpenShift SCCs and can provision arbitrary resources in their own namespaces behind quotas and existing safeguards.
* On *Managed OpenShift*, the cluster belongs to the customer, who can deploy whatever they like on it, so runner isolation is not our concern.
```

docs/modules/ROOT/pages/adr/0054-vshn-managed-forgejo-runners.adoc
```diff
@@ -0,0 +1,402 @@
...
+
+include::partial$adr-meta.adoc[]
+
+[NOTE]
+.Summary
+====
+We offer per-instance Forgejo Actions runners as an optional, separately billed *AddOn*, following the Collabora model.
+Each runner is deployed in the Forgejo instance namespace for simplicity.
+Registration uses `provider-http` to call the Forgejo API directly.
+The resulting credential is written as a Kubernetes Secret, which the `forgejo-runner` Helm chart consumes.
```
tobru, line 18: Reading current discussions in sales channels, I'm wondering if we should treat "Forgejo Action Runner" as its own service and not as an addon. I see some requests for "runner only" services, while the customer brings its own Forgejo. If we build up the service from the beginning this way, we wouldn't have to change later.

tobru (changes requested):

docs/modules/ROOT/pages/adr/0054-vshn-managed-forgejo-runners.adoc
mikeshootzz: The scope of this ADR was defined as having this as an addon and not a standalone service. I will add some considerations for this but this, but from the information I got from @lieneluksika we currently do not have any concrete inquiries for this. I would keep the decision the same

Kidswiss (approved):

mdnix (approved):

PR #200, opened 2026-06-05
Add Slapper ADRs

Kidswiss:
## Summary
This ADR describes the current state of the slapper.

Additionally, move the framework 1.0 ADRs under framwork 1.0. We can later hand-pick ADRs which apply to both versions.

## Checklist


docs/modules/ROOT/pages/framework/framework-2-0/adrs/0001-api-converter.adoc
```diff
@@ -0,0 +1,85 @@
...
+
+=== kro SimpleSchema for service-specific parameters
+
+We will embed the maintainer's service-specific parameter schema as a kro `SimpleSchema` fragment under `claim.simpleSchema`. The converter expands it through `github.com/kubernetes-sigs/kro/pkg/simpleschema` into an OpenAPI v3 subtree and inlines it under `spec.parameters.service` in the generated XRD. Framework-owned fields (`plan`, `instances`, `maintenance`) sit alongside it under `spec.parameters` and are owned by the framework, not the maintainer.
+
+=== Tagged unions for variant stanzas
+
+`Renderer`, `PipelineStep` and `CredentialValue` are each modelled as a tagged union: a single discriminator field on the YAML side (`type`, `kind`, `source`), backed in Go by an `interface` (`RendererSourceSpec`, `StepSpec`, `CredentialSpec`) and a per-variant struct. Decoding/encoding goes through generic `decodeTagged` / `encodeTagged` helpers. New variants are added by defining a struct, implementing the variant interface, and registering a factory. This keeps the YAML free of `kind`-specific subtrees while staying strongly typed Go-side.
+
+=== Pipeline renderer registry with dummy placeholders
```
Kidswiss, line 52: Not sure if we should have this in the ADR, as the dummies will be removed later. But as of now it might be helpful. Any opinions?

docs/modules/ROOT/pages/framework/framework-2-0/adrs/0001-api-converter.adoc
```diff
@@ -0,0 +1,85 @@
...
+:adr_date:      2026-06-05
+:adr_upd_date:  2026-06-05
+:adr_status:    draft
+:adr_tags:      framework, framework-2-0, converter, crossplane
+
+include::partial$adr-meta.adoc[]
+
+== Summary
+
+Introduce `slap`, a small Go CLI that converts a single maintainer-authored `ServiceBundle` YAML into a Crossplane v2 `CompositeResourceDefinition` and a Crossplane `Composition`, replacing the current monorepo function approach.
```
zugao, line 15: 😄

docs/modules/ROOT/pages/framework/framework-2-0/adrs/0001-api-converter.adoc
```diff
@@ -0,0 +1,85 @@
...
+
+== Decision
+
+We will introduce a Go CLI `slap` (package: `github.com/vshn/slapper`) that takes a single `ServiceBundle` YAML and emits an XRD plus a Composition under `xpkg/`.
+
+The implementation is anchored on the following decisions.
+
+=== ServiceBundle as the single declarative input
+
+We define one Go-typed YAML schema (`pkg/servicebundle.ServiceBundle`) with a small set of top-level stanzas: `meta`, `claim`, `renderer`, `pipeline`, `credentials`, `plans`. Maintainers author this once; the converter derives everything else.
```
zugao, line 40: Good, we make sure the top level objects are as we (FE) defined them.

docs/modules/ROOT/pages/framework/framework-2-0/adrs/0001-api-converter.adoc
zugao: I am not sure if this whole document should really be considered an ADR at this point. I would do it a bit differently, I would have a document were we actively update it and add features to it the more me work on the new framework. Later we can derive ADRs out of this document, these would be almost final decisions based on the initial documentation and PoCs. The main reason behind this thinking is that, I expect this document to be overwritten/updated multiple times in order to arrive to a first initial polished solution.

docs/modules/ROOT/pages/framework/framework-2-0/adrs/0001-api-converter.adoc
```diff
@@ -0,0 +1,85 @@
...
+
+=== Pipeline renderer registry with dummy placeholders
+
+`pkg/converter/pipeline` ships a `Register` / `Get` registry keyed on `PipelineStepKind`. Until the stdlib lands, `registerDefaults()` populates the registry with a `dummyRenderer` for each built-in kind that emits a `function-kcl` step writing a placeholder `ConfigMap`. The `customRenderer` is permanent and passes the maintainer's custom function reference through.
+
+=== Custom function name derived from the OCI reference
+
+`CustomStep.Function` is a `FuncRef { Name, VersionConstraint }`. The `functionRef.name` written into the Composition is *derived* from `FuncRef.Name` using a best-guess heuristic: take the last non-empty path segment, then strip any digest (`@sha256:…`) and tag (`:vX.Y.Z`). For built-in step kinds, the dummy renderers hardcode `function-kcl` and ignore derivation. The stdlib, when it lands, will replace these per-kind renderers wholesale and will not rely on derivation.
```
zugao: Here is a bit too much implementation details for a first version in my opinion, but I am ok to keep it.

docs/modules/ROOT/pages/framework/framework-2-0/adrs/0001-api-converter.adoc
```diff
@@ -0,0 +1,85 @@
...
+
+The XRD is emitted with `apiVersion: apiextensions.crossplane.io/v2`, `scope: Namespaced`, no `claimNames`, no `offered` flag — the v2 model is XR-only.
+
+
+== Consequences
+
+Positive:
+
+* Maintainers write one YAML per service. Kind/plural/schema drift between XRD and Composition is structurally impossible: both are derived from the same source.
+* Non-Go authors can express the service-specific schema in kro SimpleSchema rather than raw OpenAPI v3.
```
zugao, line 71: we should allow SM to be able to define these in different formats, so we should prepare the groundwork for that even though we would do the first proper PoC with SimpleSchema.

zugao (commented):
A lot of assumptions are being made on how the framework would evolve thus I expect this document will evolve a lot during the next months. As I said in a comment below I prefer this not to be an ADR, there are other approaches such as Solution Architecture Documents (SAD) or Architecture Design Document (ADD) that we might work on first before we have a proper flow in framework 2.0 and start introducing ADRs.

docs/modules/ROOT/pages/framework/framework-2-0/adrs/0001-api-converter.adoc
TheBigLee: I agree with @zugao This will most likely evolve over time as we go

TheBigLee (commented):
The whole text is too heavy on AI slang for my taste. It makes reading and understanding it harder than necessary as it uses a few termologies that are not that common in natural language

docs/modules/ROOT/pages/framework/framework-2-0/adrs/0001-api-converter.adoc
Kidswiss: I'm not sure if it makes sense to provide multiple ways to define this. It will make things more complicated for us and the SM.

docs/modules/ROOT/pages/framework/framework-2-0/adrs/0001-api-converter.adoc
Kidswiss: Yeah we will have to keep this up to date with all the implementations.

Not sure why it's not an ADR? ADR stands for Architecture Decision Record, this document contains architecture and designs of the current implementation and their decisions. It's also marked as draft, so it's not really final and is supposed to evolve.

docs/modules/ROOT/pages/framework/framework-2-0/adrs/0001-api-converter.adoc
Kidswiss: let's keep it for now, we can clean later.

mikeshootzz: This section in particular is quite hard to read. I would cut out a few inline snippets and make the wording a bit more "simpler"

docs/modules/ROOT/pages/framework/framework-2-0/adrs/0001-api-converter.adoc
zugao: What if we have a document for a macro view of the new architecture. I see ADRs more for a micro view that goes into details. But in the end it does not matter that much, it's important that we have it, I just dislike this format. I believe ADRs are strong once you have a good foundation and you come up with solutions as to why to change  something in your current project.

docs/modules/ROOT/pages/framework/framework-2-0/adrs/0001-api-converter.adoc
```diff
@@ -0,0 +1,85 @@
...
+
+=== ServiceBundle as the single declarative input
+
+We define one Go-typed YAML schema (`pkg/servicebundle.ServiceBundle`) with a small set of top-level stanzas: `meta`, `claim`, `renderer`, `pipeline`, `credentials`, `plans`. Maintainers author this once; the converter derives everything else.
+
+=== kro SimpleSchema for service-specific parameters
+
+We will embed the maintainer's service-specific parameter schema as a kro `SimpleSchema` fragment under `claim.simpleSchema`. The converter expands it through `github.com/kubernetes-sigs/kro/pkg/simpleschema` into an OpenAPI v3 subtree and inlines it under `spec.parameters.service` in the generated XRD. Framework-owned fields (`plan`, `instances`, `maintenance`) sit alongside it under `spec.parameters` and are owned by the framework, not the maintainer.
+
+=== Tagged unions for variant stanzas
```
mikeshootzz: As discussed: a link here would help. You learn some new terms every day :D

mikeshootzz (changes requested):
I agree with this ADR, but the wording needs some love. Generally speaking, I'd reduce the symbol soup and make the wording a bit "easier"

docs/modules/ROOT/pages/framework/framework-2-0/adrs/0001-api-converter.adoc
Kidswiss: I can brainstorm a different format.

docs/modules/ROOT/pages/framework/framework-2-0/adrs/0001-api-converter.adoc
Kidswiss: Let's cut this thing out completely. It's probably too implementation specific.

mikeshootzz (approved):

docs/modules/ROOT/pages/framework/framework-2-0/adrs/0001-api-converter.adoc
zugao: you sure you want to make available only one way the SM can define their API? We allow many converters but API we allow only in one format to be defined.

Kidswiss: I'd like not to overcomplicate this aspect at this point. If we ever have the need for more felxibility in the future, we can re-evaluate adding more formats.

The other converters cover a much broader aspect, since they only capture an intent (e.g. "i want tls") where the implementation can be different depending on the environment or customer. But the API will always be an XRD (or more specific openapi v3).

zugao (approved):

